PR #1, opened 2016-11-22
move code

jquense:
Moving code over from fb/react

I’m not entirely sure how to handle licensing in this case..

### DIFF
--- a/src/utils/ChildMapping.js
+++ b/src/utils/ChildMapping.js
@@ -53,7 +53,7 @@ export function mergeChildMappings(prev, next) {
   let nextKeysPending = {};
 
   let pendingKeys = [];
-  for (var prevKey in prev) {
+  for (let prevKey in prev) {
     if (next.hasOwnProperty(prevKey)) {
       if (pendingKeys.length) {
         nextKeysPending[prevKey] = pendingKeys;
@@ -66,12 +66,12 @@ export function mergeChildMappings(prev, next) {
 
   let i;
   let childMapping = {};
-  for (var nextKey in next) {
+  for (let nextKey in next) {
     if (nextKeysPending.hasOwnProperty(nextKey)) {
       for (i = 0; i < nextKeysPending[nextKey].length; i++) {
         let pendingNextKey = nextKeysPending[nextKey][i];
         childMapping[nextKeysPending[nextKey][i]] = getValueForKey(
-          pendingNextKey
+          pendingNextKey,
         );
       }
     }
--- a/src/utils/ChildMapping.js
+++ b/src/utils/ChildMapping.js
@@ -1,0 +1,87 @@
+import { Children } from 'react';
+
+
+/**
+ * Given `this.props.children`, return an object mapping key to child.
+ *
+ * @param {*} children `this.props.children`
+ * @return {object} Mapping of key to child
+ */
+export function getChildMapping(children) {
+  if (!children) {
+    return children;
+  }
+  let result = {};
+  Children.map(children, child => child)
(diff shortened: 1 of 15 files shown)


jquense:
There where a few (a lot) of internal api usage, most was easy enough to fix

- `dom-helpers` instead of CSSCore, we use it on react-bootstrap so probably a safe swap.
- Remove the transition event code branch entirely (it was already deprecated) and depend on timeouts. we can always use the dom-helpers version if we want to add it back
- Do some really hacky stuff with Children to avoid needing `flattenChildren`

jquense:
cc @gaearon

gaearon:
cc @zpao, how should we handle LICENSE/PATENTS files and copyright headers?

taion:
Wasn't there a question a bit ago about support for raw numbers between CSSCore and dom-helpers? Or is that not applicable here?

gaearon:
>Remove the transition event code branch entirely (it was already deprecated) and depend on timeouts. we can always use the dom-helpers version if we want to add it back

Can we keep it for one version? So that initial release is functionally identical to that provided by React 15. Then we can cut 2.x that drops it.

jquense:
> Can we keep it for one version? So that initial release is functionally identical to that provided by React 15. Then we can cut 2.x that drops it.

sure, no problem

jquense:
> Wasn't there a question a bit ago about support for raw numbers between CSSCore and dom-helpers? Or is that not applicable here?

yes but in case its just using the className utils, not styles

src/utils/ChildMapping.js
```diff
@@ -0,0 +1,87 @@
...
+ */
+export function getChildMapping(children) {
+  if (!children) {
+    return children;
+  }
+  let result = {};
+  Children.map(children, child => child)
+    .forEach((child) => {
+      result[child.key] = child;
+    });
```
gaearon: Does something like `Children.toArray(children)` help? I don't actually know this part of the code well.

jquense: I don't think so, or at least not fully. 

from what I gather, `flattenChildren` will return  object where the keys are the react key/data-id and values the corresponding children. my code above is attempting to emulate it by running it through `Children.map` which flattens and  adds keys to unkeyed elements. The forEach turns it into a hash.

The double iteration isn't great, but also we lose the extra debug info that `flattenChildren` gave, pointing out duplicates keys.

gaearon:
Let's do it 👍

jquense:
ok last bit added 👍

jquense:
ok released! If all want to get the word out for folks to try it out, to make sure I didn't break anything in the move.

gaearon:
Do you mind forking the docs too?

gaearon:
Docs have a separate license: https://github.com/facebook/react/blob/master/LICENSE-docs

jquense:
yes eventually, but a bit short on time right now! For now though, i don't think linking to the original docs places should be licensing issue right?

gaearon:
Sure it's fine. I just mean in case you want to make API changes or anything.

Also the docs refer to `React.addons` so beginners will likely not realize they need to use a different package.

jquense:
yeah its not a great solution in the medium to long term. We'll get some proper docs up on a gh-pages soon as I have some time to fork and extract the current ones

gaearon:
I think it's also fine just to copy & paste this on README and leave it on GitHub.
You'll miss out on our custom line highlighting but otherwise it's okay.

jquense:
ha true, I forgot it's all markdown :P

PR #4, opened 2016-12-10
Allow TransitionChild to passprops. Useful for component wrapper that…

khankuan:
… clone children and modify props, new props should be passed down to child.

Reference: https://github.com/facebook/react/pull/8522

### DIFF
--- a/src/CSSTransitionGroupChild.js
+++ b/src/CSSTransitionGroupChild.js
@@ -158,7 +158,13 @@ class CSSTransitionGroupChild extends React.Component {
 
   render() {
     const props = { ...this.props };
-    Object.keys(propTypes).forEach(key => delete props[key]);
+    delete props.name;
+    delete props.appear;
+    delete props.enter;
+    delete props.leave;
+    delete props.appearTimeout;
+    delete props.enterTimeout;
+    delete props.leaveTimeout;
     return React.cloneElement(React.Children.only(this.props.children), props);
   }
 }
--- a/src/CSSTransitionGroupChild.js
+++ b/src/CSSTransitionGroupChild.js
@@ -157,7 +157,9 @@ class CSSTransitionGroupChild extends React.Component {
   }
 
   render() {
-    return React.Children.only(this.props.children);
+    const props = { ...this.props };
+    Object.keys(propTypes).forEach(key => delete props[key]);
+    return React.cloneElement(React.Children.only(this.props.children), props);
   }
 }
 
--- a/package.json
+++ b/package.json
@@ -50,6 +50,7 @@
     "babel-plugin-add-module-exports": "^0.2.1",
     "babel-plugin-dev-expression": "^0.2.1",
     "babel-preset-latest": "^6.16.0",
+    "babel-preset-react": "^6.16.0",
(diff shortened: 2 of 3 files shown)


src/CSSTransitionGroupChild.js
```diff
@@ -157,7 +157,9 @@ class CSSTransitionGroupChild extends React.Component {
   }
 
   render() {
-    return React.Children.only(this.props.children);
+    const props = { ...this.props };
+    Object.keys(propTypes).forEach(key => delete props[key]);
```
taion: this pattern feels a bit odd to see here – i'm afraid of people using this with the babel optimization to strip prop types and hitting breakage

jquense:
hi there, thanks for the PR. I'm not sure I understand why this is needed. The user can already add any props to the children when they create them the first time. For a wrapper that controls/adjusts children why not wrap the TransitionGroup itself instead of the Component? 

let me know if I'm misunderstanding

khankuan:
Here's the example that was unexpected:
```
<TransitionGroup {...this.props} component={List}>
   <ListItem>x</ListItem>
</TransitionGroup>
```
`ListItem` no longer has props that `List` added. Updated to remove specific props instead of using propTypes.

khankuan:
Hi, is this repo still maintained?

jquense:
is it maintained? yes, yes it is.

khankuan:
Is there anything blocking to have the PR merged? Thanks!

jquense:
I'm not sure this is the right approach... expectung the animation container to behave correctly when passing in arbitrary components (like List) seems fraught. who is to say that component will play nice. what if it doesn't render the children at all? in general it seems better for List to render a Transition group (ideally). 

@taion thoughts here?

jquense:
(sorry for the delay btw)

khankuan:
I think the original intention of the `component` prop is to allow custom wrappers instead of the default div. If List is to support `component` prop, it would look like:
```
<List component={TransitionGroup}>
   <ListItem>x</ListItem>
</List>
```
The same functionality would need to be built into other wrappers, like a `Grid`. 

I feel that its better practice for components that are most reusable to play nice rather than have dependencies above playing nice downwards. 

TransitionGroup assumed to be the wrapper hence it enforce either a div or a custom component, so it make sense that the tradeoffs of the assumption is taken care of.

Alternatively, i propose `TransitionGroup` is be more like a decorator/enhancer kind of component that wraps a parent container rather than making its own div.

taion:
@jquense Taking React-Bootstrap as an example, how would we make this work with something like `<Nav>` and `<NavItem>`?

jquense:
yeah I guess it doesn't work in the case which isn't ideal. My concern here is that it may be misleading that it would work with any Component, which isn't true if the rendered root decides to do anything odd with the children. I guess its fine to let the buyer beware tho on those things.

khankuan:
I think the goal is to reduce the cases where adding transitions via TransitionGroup fail to work well. I don't think any component is meant to work with everything, imagine `<TransitionGroup component={ReduxProvider}>`. We probably want to support as much sensible cases as possible.

jquense:
fair enough

Issue #8, opened 2017-01-10
Any better way to get refs in ReactTransitionGroup?


techctu:
I want to implement pull-to-refresh function in my list. I suppose I should use react-transition-group.

I think it's strange to implement like:

```
// Just a sample
<div onTouchMove={(e) => this.setState({headerHeight: e.touches[0].pageY - startPosition})}>
    <div style={height: this.state.headerHeight}></div>
</div>
```

I think it's strange to update state in touchMove. I don't know what's the react way. So I suppose I can get DOM node by refs and implement animation in touchMove, then call callback function in touchEnd. 

I've found I can get children's refs in ReactTransitionGroup. Just like

```
<TransitionGroup ref={(ref) => this._refs = ref}>
    <div>Hello</div>
</TransitionGroup>
```

Then I can get ```<div>Hello</div>``` by ```this._refs.refs['.$.0']```

It's quite ugly. 

Is there any better way or further plan to implement something like ```componentWillChange``` which is called after ```componentDidUpdate```?

[Any better way to get refs in ReactTransitionGroup?](https://github.com/facebook/react/issues/8734)

jquense:
there isn't a better way to get child refs right not. we need to not remove user added ones internally for that to work. in the mean time tho you can just access the dome now children directly from the group ref. `findDOMNide(this._refs).children`

gaearon:
For the future reference, you could implement this by switching TG to use callback refs.
It probably wouldn't even be a breaking change since it used to always override refs.

PR #9, opened 2017-01-20
Support refs on children

jquense:


jquense:
rel: #8

src/TransitionGroup.js
```diff
@@ -187,6 +189,11 @@ class TransitionGroup extends React.Component {
     for (let key in this.state.children) {
       let child = this.state.children[key];
       if (child) {
+        let isCallbackRef = typeof child.ref === 'function';
+        warning(!child.ref || isCallbackRef,
```
taion: why not just `typeof child.ref !== 'string'`? can `ref` be an empty string?

jquense: ha, duh. not sure why I was making that so complicated.

PR #14, opened 2017-02-08
Add npm run lint

koba04:
This PR is for adding `npm run lint`, which is to lint code in `src` and `test` directories.
In addition to that, I've fixed a lint error in `test`.

### DIFF
--- a/test/CSSTransitionGroup-test.js
+++ b/test/CSSTransitionGroup-test.js
@@ -298,9 +298,7 @@ describe('CSSTransitionGroup', () => {
       state = { showChild: true };
 
       componentDidMount() {
-        setTimeout(() => {
-          this.setState({ showChild: false });
-        });
+        this.setState({ showChild: false });
       }
 
       render() {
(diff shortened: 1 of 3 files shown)


test/CSSTransitionGroup-test.js
```diff
@@ -298,7 +298,9 @@ describe('CSSTransitionGroup', () => {
       state = { showChild: true };
 
       componentDidMount() {
-        this.setState({ showChild: false });
+        setTimeout(() => {
+          this.setState({ showChild: false });
+        });
```
jquense: ?

koba04: This diff is for to fix `react/no-did-mount-set-state`...
I guess there are better ways...

jquense: lets just disable that rule (i have it disabled in a branch actually)

koba04: Disabled `react/no-did-mount-set-state`

jquense:
thanks!

koba04:
Thanks!

Issue #12, opened 2017-02-04
Broken animation finishing after unmounting


xStrom:
If `ReactTransitionGroup` is unmounted before all its animations have finished a javascript error is thrown complaining about reading properties of undefined.

Specifically the callbacks `_handleDoneAppearing`, ` _handleDoneEntering`, `_handleDoneLeaving` have:

```javascript
let component = this.childRefs[key];
// The following line (and its analogues) will throw with:
// Uncaught TypeError: Cannot read property 'componentDidLeave' of undefined
if (component.componentDidLeave) {
```

This bug is especially common with long running animations, however it can also happen with short animations if the timing is right.

PR #15, opened 2017-02-09
Fix TransitionGroup error on quick toggle of components

scyrizales:
I had this pull request to address this issue: https://github.com/reactjs/react-transition-group/issues/12

### DIFF
--- a/src/TransitionGroup.js
+++ b/src/TransitionGroup.js
@@ -101,7 +101,7 @@ class TransitionGroup extends React.Component {
 
   _handleDoneAppearing = (key) => {
     let component = this.childRefs[key];
-    if (component.componentDidAppear) {
+    if (component && component.componentDidAppear) {
       component.componentDidAppear();
     }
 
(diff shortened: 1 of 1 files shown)


xStrom:
How sure are we that returning is the correct behavior? By just immediately returning `currentlyTransitioningKeys` will remain inconsistent.

Previous attempts ([#6710](https://github.com/facebook/react/pull/6710), [#6745](https://github.com/facebook/react/pull/6745)) only guarded execution of code that directly depends on `component` being defined, but still executed the other parts of the function like `currentlyTransitioningKeys` cleanup.

Unless there's a good case for immediately returning, I think just guarding the `component` usage is a better solution.

jquense:
I agree with @xStrom, we shouln't return early here, but guard the actual callback so the rest of the state can be properly cleaned up

scyrizales:
@xStrom I have just updated the PR

xStrom:
Looks good, I think we should merge this.

PR #24, opened 2017-04-03
Refactor API

jquense:
This is an initial stab at a new api for these components.  The goals here are:

### Goals:

- Provide better low level hooks, while also avoiding the wonky `childFactory` pattern.
- Write the TransitionGroup component in terms of a more primitive `Transition` component
- Separate out the mount/transition state machine from implementations (like CSS)
- Simplify the user facing API and providing something a bit more idiomatic React.

### Changes

- Adds a new component `Transition` that handles the animation of a single child. 
- Simplify `TransitionGroup`, deferring a lot of state management to `Transition` which are now required as children of `TransitionGroup`
- Remove CSSTransitionGroup, replacing it's functionality with the new `CSSTransition` component, which can be used with `TransitionGroup`

These changes effectively move the transition definitions to the _list items_ instead of list. It's a bit more verbose (easily encapsulated in a component, however), but it provides greater flexibility. For instance you can now easily mix and match transition types for list items, some could Fade, and others slide in. It also provides an opportunity to reuse and share transitions a bit easier. For instance, the react-bootstrap [`Fade`](https://react-bootstrap.github.io/components.html#transitions-fade) component can be implemented in terms of a Transition, and used by itself to animate a single component, or with `TransitionGroup` to animate lists of things.

#### Example

```jsx
<TransitionGroup>
  {listItems.map(item => (
    <CSSTransition 
       key={item.id} 
       timeout={500}
       classNames="fade"
     >
      <div>{item.name}</div>
    </CSSTransition>
  ))}
</TransitionGroup>
```
with the following css
```css
  .fade-enter,
  .fade-exit {
    transition: opacity 500ms;
  }
  .fade-enter,
  .fade-exit-active {
    opacity: 0;
  }
  .fade-exit,
  .fade-enter-active {
    opacity: 1
  }
```

The removal of the various static method hooks also means it's much easier to encapsulate animations in Components. So we could write the above as 

```
function Fade() {
  return (
    <CSSTransition
      {...this.props}
      classNames="fade"
      timeout: 500,
    />
  );
}
```

### Implications and Concerns

This model complicates concept of `appear` transitions. Since it's the `Transition` are always being mounted in a `TransitionGroup` as items enter, they are always "appearing". Because the list items are controlling the transition, they don't know whether they are appearing or entering in relation to the parent list. This is somewhat solvable, but a bit ugly. **I'd prefer to remove the concept of "appear" from the items** in the context of a TransitionGroup, instead maybe just animation the TransitionGroup component itself, for appear animations

```js
<FadeOnAppearTransition>
  <TransitionGroup>
     {items.map(item => (
       <SlideTransition key={item.id}><div>{item}<div></SlideTransition>
     ))}
  </TransitionGroup>
<FadeOnAppearTransition>
```

Not sure. **I'd love some feedback from folks on when and if they use appear animations as separate from the enter animations**

I still need to figure out how the API will play with the TransitionGroup component, i do still want the shortcut apis for defining timeouts and enabled/disabled animations in one spot up top.

cc @taion (I moved Transition from react-overlays btw)


jquense:
`appear` is weird to handle in this api...I don't love the awkwardness of it

BowserKingKoopa:
I'd prefer if transitions, especially transitioning out (since you can do transitions in now), was part of the Component life cycle so a component could handle its own transitions.

jquense:
@BowserKingKoopa ideally that would be great. Unfortunately there isn't any real way to do that, since unmounting is not asynchronous, and we cannot do it over time, which is part of why this s so complicated to do in React in the first place :) 

The idea tho behind this Api is that you can get pretty close to that, by defining your own Transition component that is in charge of doing the actual transition, and it exposes a lot of hooks for each transition state change.

todorone:
I guess it's a good direction. Reusable animations are a big plus. And separating appear from enter animations seems not necessary.

PR #25, opened 2017-04-07
fixed ios 8.0 not have Object.assign

GoToBoy:
add a _.extend to fixed when was in ios 8.0

### DIFF
--- a/src/CSSTransitionGroup.js
+++ b/src/CSSTransitionGroup.js
@@ -2,7 +2,7 @@ import React from 'react';
 
 import TransitionGroup from './TransitionGroup';
 import CSSTransitionGroupChild from './CSSTransitionGroupChild';
-import { nameShape, transitionTimeout } from './utils/PropTypes';
+import { nameShape, transitionTimeout, _extends } from './utils/PropTypes';
 
 const propTypes = {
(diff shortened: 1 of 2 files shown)


oliviertassinari:
💯  for fixing that issue. It's also breaking support with IE11. React used to take care of that issue. For instance here is the output of `react/lib/ReactTransitionGroup`:
```js
'use strict';

var _assign = require('object-assign');

// ...

var props = _assign({}, this.props);
```

oliviertassinari:
@GoToBoy What about using `https://www.npmjs.com/package/babel-plugin-transform-object-assign` instead?

oliviertassinari:
@jquense I think that you want to have a look at it as a regression with the deprecated https://www.npmjs.com/package/react-addons-transition-group package.

jquense:
Thanks y'all, yes we want to fix this but not with a self included polyfill. Let's use the Babel plugin that handles this automatically thanks!

jquense:
And sorry for the oversight!

kenfehling:
Also came across this issue. As per @oliviertassinari's suggestion, I've opened a pull request using babel-plugin-transform-object-assign: https://github.com/reactjs/react-transition-group/pull/35

PR #28, opened 2017-04-10
Get PropTypes from prop-types package

felixhorro:
React.PropTypes is deprecated as of React v15.5. Changing to read PropTypes from prop-types package, as [seen on React docs](https://facebook.github.io/react/docs/typechecking-with-proptypes.html).

### DIFF
--- a/package.json
+++ b/package.json
@@ -44,6 +44,7 @@
   "dependencies": {
     "chain-function": "^1.0.0",
     "dom-helpers": "^3.2.0",
+    "prop-types": "^15.5.6",
     "warning": "^3.0.0"
   },
   "devDependencies": {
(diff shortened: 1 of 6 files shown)


Guria:
Would be great to see here extending peer deps to support "react@^16.0.0-0" as well.

taion:
That would be a separate PR.

package.json
```diff
@@ -63,6 +63,7 @@
     "eslint-plugin-jsx-a11y": "^4.0.0",
     "eslint-plugin-react": "^6.10.0",
     "jest-cli": "^19.0.2",
+    "prop-types": "^15.5.6",
```
taion: this needs to be a normal dependency

jquense: GOOD CATCH

felixhorro: Ok, nice point!

Will commit.

liady:
@taion any news on this PR?

PR #31, opened 2017-04-12
Check that the component exists on leave

darcyadams:
When using transition group inside a component that is unmounted using `React.unmountComponentAtNode`, the `component` object seems to get destroyed. This cuases an uncaught error: 
```
null is not an object (evaluating 'component.componentWillLeave')
```
For me, this really is only an issue in tests, where I'm ensuring `componentDidLeave` calls a callback method.

I figured it is such a small change, it would be worth fixing, even though it's a bit of an edge case.

Thanks for your time maintaining this module. We make extensive use of it in our applications.


frankwallis:
I think this is the same issue as #55

Issue #29, opened 2017-04-11
CSSTransitionGroup child ref function called with CSSTransitionGroupChild


jliuhtonen:
**Do you want to request a *feature* or report a *bug*?**

Bug

**What is the current behavior?**

`ref` function given to a `CSSTransitionGroup` child is called with both the child element refence and the CSSTransitionGroupChild component instance

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem via https://jsfiddle.net or similar (template: https://jsfiddle.net/1Lb2qk3v/).**

See console in https://jsfiddle.net/10sjj3so/6/

**What is the expected behavior?**

`ref` should only be called for the child given to `CSSTransitionGroup`, not the synthetic `CSSTransitionGroupChild` instance that is an implementation detail of `CSSTransitionGroup`

**Which versions, and which browser / OS are affected by this issue? Did this work in previous versions?**

All browsers. Worked in React addons version.

rayhatfield:
This is causing a problem in my project too.

karl:
We've just spotted this as well.

String refs seem to work correctly (as in they reference the child element and not a `CSSTransitionGroupChild`). Although they work, using a string ref causes a warning saying they are not supported and will be ignored.

I'm not sure what the correct approach is at the moment, as callback refs seem to be broken, but string refs are supposedly not supported! Any advice would be great.

karl:
We've worked around this for the moment by moving back to `react-addons-css-transition-group`version `15.5.2` which works correctly with string refs without a warning.

PR #39, opened 2017-04-19
Proposed fix for #29

aligon:
This was hijacking the child ref onto the `CSSTransitionGroupChild`, this fix for #29 leaves the ref on child and lets the `TransitionGroup` add a ref to the wrapped child produced from `childFactory()`.  (assumption: `childFactory()` always wraps the child) An improvement would to detect if the child was wrapped or not and chain appropriately.

### DIFF
--- a/src/TransitionGroup.js
+++ b/src/TransitionGroup.js
@@ -204,11 +204,9 @@ class TransitionGroup extends React.Component {
           this.props.childFactory(child),
           {
             key,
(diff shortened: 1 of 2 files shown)


aligon:
@jquense I added some tests, and went ahead and made the change to intelligently chain the refs or not if the childFactory creates a wrapper.

jsg2021:
🎉

PR #40, opened 2017-04-20
Fix import/require statements

jameslnewell:
because `react-transition-group/TransitionGroup` doesn't exist, it exists at `react-transition-group/lib/TransitionGroup`. But `TransitionGroup` is exported from the package main so we don't need to import from a specific file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,9 +15,9 @@ The [`TransitionGroup`](#transitiongroup) add-on component is a low-level API fo
 **Importing**
 
 ```javascript
-import CSSTransitionGroup from 'react-transition-group/CSSTransitionGroup' // ES6
+import {CSSTransitionGroup} from 'react-transition-group' // ES6
 
-var CSSTransitionGroup = require('react-transition-group/CSSTransitionGroup') // ES5 with npm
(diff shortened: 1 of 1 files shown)


jquense:
Hi there thanks for the PR. The files are in lib on github, but they are where the docs say they are in the npm package. We want to keep the current instructions over the main import as it encourages only importing the files needed and helps keep bundles as lean as possible

jameslnewell:
Aren't they in `./src` on Github and then transpiled to `./lib` after running `npm run build`? But the docs suggest you'll find them in the root, not the `./lib`?

I was getting `undefined` after copy pasting the imports from the docs, but importing from the package main fixed it. The app I'm working Typescript... I'm not familiar enough with Typescript to know whether that would make a difference.

jquense:
They are transpired to lib, but then only the lib directory is published to npm. You can poke around the package on unpkg.com if your curious

tkrotoff:
> We want to keep the current instructions over the main import as it encourages only importing the files needed and helps keep bundles as lean as possible

That's the job of the bundler (webpack, rollup...) not yours.

Issue #55, opened 2017-05-09
Ref missing when callback called while still entering/leaving on TransitionGroup


zediah:
**Do you want to request a *feature* or report a *bug*?**
Bug

**What is the current behavior?**
If, during a transition, the callback (_handleDoneEntering/leaving) is called. i.e. the transition is interrupted and the component is now leaving. When the "_handleDone[Entering|Leaving]" is called, the ref for the element leaving is null. 

In my instance, I am calling this callback during a componentWillReceiveProps.

This appears to be a consequence of changing to a function based ref (which I think is a good thing, but its caused by this). 

The reason this happens is when the component is updated via react (see 'receiveComponent' method). React checks to see if the refs have changed to determine if it needs to detach/attach refs before/after it updates the component. With the new function method - react determines that the refs have -always- changed as the function signature has changed (it's an anonymous function generated each time). This means that react will detach the ref - after that, the component is updated and the callback is called, the ref is null, the "handleDone" method attempts to call a method on a null component from "this._childRefs[<leavingKey>]".

Previously when using a string ref, it would see that the string is not different and would not require a change - thus leaving the ref in place.

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem via https://jsfiddle.net or similar (template: https://jsfiddle.net/1Lb2qk3v/).**

I created a fiddle:
https://jsfiddle.net/tzLojs6n/
Hit the 'add' button once, then hit it again within 3 seconds.

**What is the expected behavior?**
The ref exists and that handleDone does not throw an error.

**Which versions, and which browser / OS are affected by this issue? Did this work in previous versions?**
This did work in previous versions where a string ref was used (see the original one still included in React-addons).

I've been able to fix this by keeping a reference to the function generated per child key. Thus when react compares the refs, it sees they are the same and it works. Given the idea that the child at a given key should always be the same component (and all it does is set the same key anyway) - I believe this is a safe thing to do.

Should I create a pull request for this change?

PR #56, opened 2017-05-09
Proposed fix for #55

zediah:
Fix issues as in issue #55 

The only caveat is that I was unable to work out a way to maintain the same function ref when the child.ref is set. At the moment it'll just do it the usual way...

This fixes my situation as I don't often have ref's on the child (always wrapped).

### DIFF
--- a/src/TransitionGroup.js
+++ b/src/TransitionGroup.js
@@ -25,6 +25,12 @@ class TransitionGroup extends React.Component {
     super(props, context);
 
     this.childRefs = Object.create(null);
+    // Maintain a list of our childrens 'ref' functions
+    // This is to prevent react from treating each ref as a new ref
+    // causing the ref to be detached for each render 
+    // (this causes an error if the callback is 
+    // called during an existing onEnter/onLeave/onAppear) 
+    this.childRefFuncs = Object.create(null);
 
     this.state = {
       children: getChildMapping(props.children),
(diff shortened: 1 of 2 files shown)


jquense:
Hi there, thanks a lot for the PR. I'll admit that I'm not the most familar with the logic react uses here, but this feels "hacky" to me, I'd expect that regardless of the function identity the ref should be called with the instance (after being `null`ed out) not just broken as in our case...

@gaearon do you have any insight on this? I realize that refs are called with `null` but if the ref function changes should it go `inst -> null -> inst` again?

gaearon:
>but if the ref function changes should it go inst -> null -> inst again?

Yes, it's expected to first give you `null` before the update, and then give you the `inst` again after the update. So I think the right fix would be to ensure that any user-specified callbacks run after the DOM is flushed and refs are ready.

jquense:
Is that feasible with the public API? I can imagine wrapping callbacks in like a `forceUpdate(callback)` but that seems iffy? In the past i would just require '`lib/ReactUpdates.asap` directly ;)

gaearon:
Hmm. I don't remember how TransitionGroup works, but are `componentDidMount` and `componentDidUpdate` hooks not sufficient? They fire after refs are attached.

zediah:
My approach to this problem was that, it worked before because the refs stayed the same - so I'd replicate that. If there's a better way, I'm definitely up for it!

I agree with you having to store the function for each key seems a bit hacky, but I'd argue that this is a result of a different 'accepted' pattern now being used for refs. Previously you'd have a single string, that didn't change unless well...it did change. Now the 'pattern' is to declare an anonymous function each time - thus making it a 'new' ref, making react detach/reattach each time. Is there an advantage to this I'm not aware of? Was it a conscious decision to make this the new norm?

To me it seems a bit of an anti pattern as it'll make react do (a little bit) more work than is required and as shown here - the ref isn't available when it's expected to be. I'd also argue that the ref function in 99% of cases wouldn't change, so why are we re-creating it? (these are genuine question, I'm curious!)

@gaearon Would that mean to suggest deferring the callback call until componentDidUpdate in either the library code or my code? (or both?)

jquense:
should be fixed by #72

PR #57, opened 2017-05-09
Add loose-envify browserify transform

mridgway:
Related to https://github.com/facebook/react/issues/9641

This tells browserify to remove `process.env.NODE_ENV` conditionals during bundling.

### DIFF
--- a/package.json
+++ b/package.json
@@ -44,6 +44,7 @@
   "dependencies": {
     "chain-function": "^1.0.0",
     "dom-helpers": "^3.2.0",
+    "loose-envify": "^1.3.1",
     "prop-types": "^15.5.6",
     "warning": "^3.0.0"
   },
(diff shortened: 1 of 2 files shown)


gaearon:
(To be clear, this is important for people who don't envify globally. We used to support this in React, so it would be nice to get this in.)

jquense:
Thanks!

jquense:
Tho I don't actually think we do dev checks anywhere so a bit of a noop at the moment

mridgway:
You're right. My search on the repo found the `index.js` file in the root that had a bunch of `process.env.NODE` checks, but it doesn't seem like that's used through npm?

mridgway:
Actually, it looks like https://github.com/reactjs/react-transition-group/blob/master/src/TransitionGroup.js#L199 gets transformed into a `process.env.NODE_ENV` check.

PR #63, opened 2017-05-19
Add installation instructions to README

jmeas:
I know the title of the README happens to be the package name, but I had to dig into the `package.json` to confirm that the name lines up with what's registered on npm. These installation instructions could possibly help others more quickly install the package.


jquense:
thanks!

Issue #64, opened 2017-05-27
[Feature] Include non-minified (development) version in /dist


virgofx:
Currently, releases in the `/dist` folder contain only the production/minified version.

It would be very useful for those who organize peer dependencies and include them externally (instead of bundling everything together) to have pathways for local development and production. Specifically, being able to use project build tools to include the distribution of **React-transition-group** in dev mode alongside the current dev React setup. 

When using the minified /dist version in combination with the development React suite, errors result:
```
react.js:3287 Warning: Failed prop type: Minified exception occurred; use the non-minified dev environment for the full error message and additional helpful warnings.
    in CSSTransitionGroup (created by AppComponent)
```

Using production **react-transition-group** with production **react** suite results in no errors.

So again, when building /dist it would be nice to include 2 files:

* `react-transition-group.js` - Development/un-minified
* `react-transition-group.min.js` - Production/minified

virgofx:
Fixed by PR #66 , will be available next release.

PR #66, opened 2017-05-29
Incorporated development + production builds into /dist (#64)

virgofx:
Incorporated development + production builds into /dist (#64) and updated Readme to include CDN/External information.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,23 +22,11 @@ yarn add react-transition-group
 
 ## CDN / External
 
-To exclude the react-transition-group package from your primary build and use an external CDN when bundling, link
+Since react-transition-group is fairly small, the overhead of including the library in your application is
+negligible. However, in situations where it may be useful to benefit from an external CDN when bundling, link
 to the following CDN:
 
 [https://unpkg.com/react-transition-group/dist/react-transition-group.min.js](https://unpkg.com/react-transition-group/dist/react-transition-group.min.js)
-
-In order to exclude react-transition-group using **Webpack**, include the following configuration in the `externals`
-webpack configuration section: 
-
-```
-externals: {
-    'react-transition-group': {
-        amd: 'react-transition-group',
-        commonjs2: 'react-transition-group',
-        commonjs: 'react-transition-group',
-        root: 'ReactTransitionGroup',
-    }
-```
 
 ## High-level API: CSSTransitionGroup
 
--- a/README.md
+++ b/README.md
@@ -20,6 +20,26 @@ npm install react-transition-group --save
 yarn add react-transition-group
 ```
 
+## CDN / External
+
+To exclude the react-transition-group package from your primary build and use an external CDN when bundling, link
+to the following CDN:
+
+[https://unpkg.com/react-transition-group/dist/react-transition-group.min.js](https://unpkg.com/react-transition-group/dist/react-transition-group.min.js)
+
+In order to exclude react-transition-group using **Webpack**, include the following configuration in the `externals`
+webpack configuration section: 
+
+```
+externals: {
+    'react-transition-group': {
+        amd: 'react-transition-group',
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -20,14 +20,34 @@ npm install react-transition-group --save
...
+
 ## High-level API: CSSTransitionGroup
 
 `CSSTransitionGroup` is a high-level API based on [`TransitionGroup`](#low-level-api-transitiongroup) and is an easy way to perform CSS transitions and animations when a React component enters or leaves the DOM. It's inspired by the excellent [ng-animate](http://www.nganimate.org/) library.
 
 **Importing**
 
 ```javascript
-import CSSTransitionGroup from 'react-transition-group/CSSTransitionGroup' // ES6
+import { CSSTransitionGroup } from 'react-transition-group' // ES6
```
jquense: I'd rather show the cherry picked version here :)

README.md
```diff
@@ -20,14 +20,34 @@ npm install react-transition-group --save
 yarn add react-transition-group
 ```
 
+## CDN / External
+
+To exclude the react-transition-group package from your primary build and use an external CDN when bundling, link
+to the following CDN:
+
+[https://unpkg.com/react-transition-group/dist/react-transition-group.min.js](https://unpkg.com/react-transition-group/dist/react-transition-group.min.js)
+
```
jquense: I appreciate the extra documentation, however in this case I don't to give the external consumption version of the library such a prominent place in the docs, as it's not really recommended. In other others most folks shouldn't use the CDN version, but instead include this in a vendor bundle (maybe put on a cdn) via webpack. 

I'd just make a quick note that for folks that do need a cdn version they can use unpkg.com

virgofx: Various libraries exclude this package but depend on it. I'd argue that library components absolutely should be used from a CDN as the alternative reinforces bundling huge apps, SPA, large bundles with users not sure how to split/reuse common framework components. E.g. React, Bootstrap, etc. This is an extension of React essentially. 

Nonetheless, would you prefer I move to the bottom or remove?

README.md
virgofx: I wasn't able to get the cherry picked version to exclude properly when testing.  All documentation everywhere prefers the standard ES6 brace style for non-default components. Do you have an example webpack external config where using cherry picked version will work?

virgofx:
@jquense Thanks for the quick feedback. I'll try to update soon as I get your latest answers to my last Qs!

README.md
jquense: externalizing dependencies is balance between size, usage and geographic location. Libraries like React may make sense to consume from a CDN since it's fairly large, and fairly likely to already be in a user's cache since lots of sites use it. Smaller libraries like this are way less likely to already be cached and are so small I'd argue the latency of a new network request outweighs the benefit of making it external. Not to mention request limit bottlenecks in browsers. 

That said folks should measure and pick for their use cases, but I think generally making all libraries external isn't great default advice and I'd rather just stay out of that game :) to that end we can keep the CDN and dist build info but let's remove the webpack config details since that's more build specific

README.md
jquense: Ya you can't cherry pick from the dist builds since they are flat files (a downside of UMD builds). We really should just both options since they are all valid choices for importing. I'm fine leaving this change in. I can make the docs more detailed later

virgofx: Gotcha. Dropped the webpack build specific option above and added minor clarity. If you want to clarify the import or even better document both (Cherry picked for direct bundled or ES6 style import for those using /dist builds), fine with me.

jquense:
Thanks!

virgofx:
@jquense Thanks as well. When you cut the next release can you update the dates and link in the ChangeLog to #64 . I forgot to link that to the issue. :)

PR #68, opened 2017-06-01
Prevent `this.setState` if there is no component

WesleydeSouza:
On `_handleDoneLeaving`, prevent calling `setState` if the component is unmounted.

This was triggering this error:

> Warning: setState(...): Can only update a mounted or mounting component. This usually means you called setState() on an unmounted component. This is a no-op. Please check the code for the TransitionGroup component.

This happened on a very complex structure and I was not able to reproduce it with sample code.

It is related to #15.

### DIFF
--- a/index.js
+++ b/index.js
@@ -437,7 +437,7 @@ var TransitionGroup = function (_React$Component) {
       if (currentChildMapping && currentChildMapping.hasOwnProperty(key)) {
         // This entered again before it fully left. Add it again.
         _this.performEnter(key);
-      } else {
+      } else if (component) {
         _this.setState(function (state) {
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -437,7 +437,7 @@ var TransitionGroup = function (_React$Component) {
       if (currentChildMapping && currentChildMapping.hasOwnProperty(key)) {
         // This entered again before it fully left. Add it again.
         _this.performEnter(key);
-      } else {
+      } else if (component) {
```
jquense: this `component` isn't the TransitionGroup instance. The error would be related to TransitionGroup not being mounted when setState is called, it'd be better to check if the current component is mounted

PR #70, opened 2017-06-03
Enhancement for best practice prop types.

### LINKED ISSUE #69: [Feature] Add capability to remove prop types for extending applications
Prop types in this package are static and included in both development and production. Best practice is to use the plugin [babel-plugin-transform-react-remove-prop-types](https://github.com/oliviertassinari/babel-plugin-transform-react-remove-prop-types) with mode wrap. This allows developers to use the development/ES/imported version with prop types or use the external version that is minified fully.

virgofx:
Fixes #69: Reduces output bundles by wrapping prop types allowing for end users to further remove them using the babel plugin in combination with webpack define plugin. Production build file reduced by removing propTypes for static dist build.

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -9,6 +9,7 @@
   "plugins": [
     "dev-expression",
     "add-module-exports",
-    "transform-object-assign"
+    "transform-object-assign",
+    ["transform-react-remove-prop-types", { mode: "wrap" }]
   ],
 }
--- a/package.json
+++ b/package.json
@@ -57,6 +57,7 @@
     "babel-plugin-add-module-exports": "^0.2.1",
     "babel-plugin-dev-expression": "^0.2.1",
     "babel-plugin-transform-object-assign": "^6.22.0",
+    "babel-plugin-transform-react-remove-prop-types": "^0.4.5",
     "babel-preset-latest": "^6.24.0",
     "babel-preset-react": "^6.23.0",
     "babel-preset-stage-2": "^6.18.0",
--- a/src/CSSTransitionGroup.js
+++ b/src/CSSTransitionGroup.js
@@ -5,22 +5,6 @@ import TransitionGroup from './TransitionGroup';
 import CSSTransitionGroupChild from './CSSTransitionGroupChild';
 import { nameShape, transitionTimeout } from './utils/PropTypes';
 
-const propTypes = {
-  transitionName: nameShape.isRequired,
-
-  transitionAppear: PropTypes.bool,
-  transitionEnter: PropTypes.bool,
-  transitionLeave: PropTypes.bool,
-  transitionAppearTimeout: transitionTimeout('Appear'),
-  transitionEnterTimeout: transitionTimeout('Enter'),
-  transitionLeaveTimeout: transitionTimeout('Leave'),
-};
-
-const defaultProps = {
-  transitionAppear: false,
-  transitionEnter: true,
-  transitionLeave: true,
-};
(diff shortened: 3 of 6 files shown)


virgofx:
@jquense Let me know your thoughts. If we can get this merged and a future build cut relatively soon that'd be great. Thanks!

virgofx:
Ping @jquense

src/CSSTransitionGroup.js
```diff
@@ -53,7 +37,21 @@ class CSSTransitionGroup extends React.Component {
   }
 }
 
-CSSTransitionGroup.propTypes = propTypes;
-CSSTransitionGroup.defaultProps = defaultProps;
+CSSTransitionGroup.propTypes = {
```
jquense: I don't really like the propTypes defined down here, is the plugin really not smart enough to extract them as they were?

package.json
```diff
@@ -57,6 +57,7 @@
     "babel-plugin-add-module-exports": "^0.2.1",
     "babel-plugin-dev-expression": "^0.2.1",
     "babel-plugin-transform-object-assign": "^6.22.0",
+    "babel-plugin-transform-react-remove-prop-types": "^0.4.5",
```
jquense: why is this here? you aren't using it anywhere

virgofx: Good catch, forgot to add plugin wrap mode to `.babelrc`

src/CSSTransitionGroup.js
virgofx: The plugin traverses the AST using the default parser... I think it's Babylon under the hood. It wraps the direct node and can't reverse engineer because it would require to find all references I believe which is very complicated.

jquense: its not _that_ complicated, i've written that babel plugin before.

.babelrc
```diff
@@ -9,6 +9,7 @@
   "plugins": [
     "dev-expression",
     "add-module-exports",
-    "transform-object-assign"
+    "transform-object-assign",
+    ["transform-react-remove-prop-types", { mode: "wrap" }]
```
jquense: is the idea that wrapping will work with uglify to remove dead code?

virgofx: Yes, correct.  Sample output bundle from development build would look as follows:

```
TransitionGroup.propTypes = process.env.NODE_ENV !== "production" ? {
  component: _propTypes2.default.any,
  childFactory: _propTypes2.default.func,
  children: _propTypes2.default.node
} : {};
```

In the now production build for this app (as well as end-users who import and build production bundles using either Webpack `-p` or equivalently **Define** + **Uglify** plugins) will result in:

**After define**
```
TransitionGroup.propTypes = false ? {
  component: _propTypes2.default.any,
  childFactory: _propTypes2.default.func,
  children: _propTypes2.default.node
} : {};
```

**After uglify **
```
TransitionGroup.propTypes = {};
```

src/CSSTransitionGroup.js
virgofx: I could investigate to see whether uglify will find the dead code and resulting file size is same if you'd prefer to keep these as separate variables?

virgofx:
@jquense The uglify plugin **does** remove the dead code for prop types when defined in the original location, so I went ahead and restored those. So essentially, this saves a few hundred bytes in production now with the following changes :)

virgofx:
ping @jquense

jquense:
thanks!

PR #72, opened 2017-06-06
Alternative fix for #55

frankwallis:
This is an alternative to PR #56 addressing issue #55 which seems to have stalled.

I have used the same test @zediah provided and added another one for the 'complementary' error which happens when going in the other direction.

It takes a different approach which is to keep a separate ref for keys which are currently transitioning and use those in the callback function. It also pushes to ```keysToEnter``` when it detects that a component has been re-added while leaving, this enables performEnter to be called in componentDidUpdate after the  ref has been setup.

There may be some situations which are not accounted for but all the tests pass. All feedback is appreciated, many thanks.

### DIFF
--- a/src/TransitionGroup.js
+++ b/src/TransitionGroup.js
@@ -41,7 +41,7 @@ class TransitionGroup extends React.Component {
     let initialChildMapping = this.state.children;
     for (let key in initialChildMapping) {
       if (initialChildMapping[key]) {
-        this.performAppear(key);
+        this.performAppear(key, this.childRefs[key]);
       }
     }
   }
@@ -60,7 +60,7 @@ class TransitionGroup extends React.Component {
     for (let key in nextChildMapping) {
       let hasPrev = prevChildMapping && prevChildMapping.hasOwnProperty(key);
       if (nextChildMapping[key] && !hasPrev &&
-          !this.currentlyTransitioningKeys[key]) {
+        !this.currentlyTransitioningKeys[key]) {
         this.keysToEnter.push(key);
       }
     }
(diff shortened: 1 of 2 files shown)


src/TransitionGroup.js
```diff
@@ -87,9 +87,8 @@ class TransitionGroup extends React.Component {
   }
 
   performAppear = (key) => {
-    this.currentlyTransitioningKeys[key] = true;
-
     let component = this.childRefs[key];
+    this.currentlyTransitioningKeys[key] = component;
```
jquense: my concern with this approach is that there is now a worry that refs will be kept around after they are unmounted accidentally. If any case is missed you'll end up with a hash of objects that are being kept alive forever

frankwallis: I understand the concern, but I'm not convinced that this makes things any worse than they are at the moment. The only time we are holding on to an extra ref is in the situation when the child component does not call the callback in which case I think it is justified to keep the element alive?

I could write some tests around this , but I'm not quite clear how elements get 'accidentally' unmounted - do you mean when someone calls ```unmountComponentAtNode``` or when someone does ```element.removeChild``` directly on the DOM?

When an element does get 'accidentally' unmounted do we expect React to even call the ref function to null out the original ref?

frankwallis:
I have changed it back so that it just maintains a boolean flag for the currently transitioning keys. I have now added the component instance as a parameter to all of the transitioning methods and the correct instance is bound to the callback function before it is given to the child. I believe that now, once a ref is nulled out it will only be kept alive by the callback function itself, so once that goes out of scope it should be released.

The other changes are just reverting the code back to how it was originally, so it may be easiest to review a diff with current master (or view 'all changes'). Let me know what you think.

jquense:
awesome thanks !

PR #74, opened 2017-06-13
Reword 1.x notice

gaearon:
Goals:

* Make it less scary
* Point people who are migrating to the right README


jquense:
Perfect thanks!

Issue #73, opened 2017-06-11
[Feature]: Sequenced / Staggered Transition Support


Setsun:
## Do you want to request a *feature* or report a *bug*?
Feature

## What is the current behavior?
Currently I don't believe there is support in the library for supporting sequenced or staggered transitions. 

Example: Say I have 10 items that enter at the same time in a CSSTransitionGroup. The current behavior is that if I have a `transitionEnter` CSS class defined, all of these animate at the same time.

## What is the expected behavior? 
Following on the previous example, if I have 10 items that enter a `CSSTransitionGroup` at the same time I want to be able to specify if they stagger in or come one after another (opposed to animating at the same time). Example below for staggered transitions, the sequential one would be similar, but distinct in that transitions can only kick off after the previous one has finished.

#### Staggered Transition Example GIF
![staggered](https://user-images.githubusercontent.com/4651424/27013437-cb05e1c8-4eb1-11e7-81e6-d9f5f99c52ae.gif)

#### Ideas on implementation
- Staggered: I have a suboptimal solution implemented as a Higher-Order Component here using a `setInterval` (which is not great) here: https://github.com/Setsun/react-choreography/blob/master/src/transitions/sequenced/StaggeredTransitionDecorator.js

- Sequential: I think this could be implemented using `transitionend` events, where the next transitioning component mounts after the previous component fires off its `transitionend` event.

Thoughts? I could spin up a PR in my free time to get a quick proof of concept since I think this would be useful for others, but I also know the API is in-flux in this PR: https://github.com/reactjs/react-transition-group/pull/24

Setsun:
For anyone interested, I've opened a PR here: https://github.com/reactjs/react-transition-group/pull/78

silvenon:
For posterity: see #78 and #340 for context about why we decided to not implement this feature.

PR #78, opened 2017-06-20
Staggered / Sequential Transitions 

Setsun:
Addresses https://github.com/reactjs/react-transition-group/issues/73 for adding Staggered / Sequenced transition support.

The approach here is to allow a new `staggerTime` property on the `TransitionGroup` component.

This `staggerTime` is used to apply an inline style object containing a `transition-delay` CSS property to each React Child passed down to the `TransitionGroup`, which is calculated with the following formula.

```
const style = {
   transitionDelay: `${this.props.staggerTime * (childIndex + 1)}ms`
}
```

### Storybook Demo:
transition-delay-css

I may adjust this PR to support passing down a staggerTime to a Transition child as well, there are some other things that could potentially be added.

There are probably some things I am missing context on, feedback whenever anyone gets the chance would be much appreciated.

cc/ @jquense


src/TransitionGroup.js
```diff
@@ -37,12 +37,31 @@ const propTypes = {
...
 
 const defaultProps = {
   component: 'div',
 };
 
+function childMapperFn(child, index) {
+  const { staggerTime } = this.props;
+  const onExited = () => this.handleExited(child.key);
+  const style = staggerTime > 0 ? {
+    transitionDelay: `${staggerTime * (index + 1)}ms`
```
jquense: This shouldn't go here, since it's css specific and this component is intended to be a generic component. I believe you should actually be able to implement this with the API as it is with minimal fuss. Your Transition children can accept a 'delay' property, and either compute it your as you map over the children or wrap TransitionGroup to inject it with similar logic.

```
<TransitionGroup>
  {items.map((item, idx) => 
    <StaggeredFade delay={staggerItem * (idx + 1)}>{item}</StaggeredFade>
   )}
</TransitionGroup>
```

Setsun: Ah, I see your point. I'll see if I can re-work this without CSS. Thanks for the quick response.

Setsun:
@jquense I've retooled the PR as follows, which should support several scenarios.

The `Transition` component now has a `delay` property, which has the exact same shape as the `timeout` property. This means we can define separate `enter` and `exit` delays if we so desired. Otherwise it will apply the same delay to both. This allows us to manually set delays for individual children if the need arises.

The `TransitionGroup` can also set a global `delay` if needed and has a `stagger` property which is for convenience. You could alternatively manually calculate the delays for each child if you so wanted. However if `stagger` is true, we calculate all of the delays for you if you pass down a `delay` directly to the `TransitionGroup` as well.

The difference between a staggered / sequential animation has little difference with this API. To get a sequential animation over a staggered one, one would pass down a `delay` property that is equal to the `timeout` property to the `TransitionGroup` and set `stagger` to true.

## Storybook Examples

#### Staggered Animations
staggered-r

#### Sequential Animations
sequential-r

#### Single Delayed Animation
delayed-r

Setsun:
Also some examples on what this may look like in action:

## Example Usage
```
// Global Delay
<TransitionGroup
  timeout={1500}
  delay={300}>
  {[...new Array(10).keys()].map(() => (
     <CSSFadeTransition />
  ));
</TransitionGroup>

// Staggered Transition
<TransitionGroup
  timeout={1500}
  delay={300}
  stagger={true}>
  {[...new Array(10).keys()].map(() => (
     <CSSFadeTransition />
  ));
</TransitionGroup>

// Sequential Transition
<TransitionGroup
  timeout={1500}
  delay={1500}
  stagger={true}>
  {[...new Array(10).keys()].map(() => (
     <CSSFadeTransition />
  ));
</TransitionGroup>
```

Setsun:
@jquense any update on this? The API seems sound to me, but let me know if there are any remaining questions / work to be done on this.

jquense:
Hey there :wave: I appreciate the work you've put in here. My major concern is that this adds a lot of complexity to already complex components. I'm also uncomfortable with the bluring of some component boundaries. For instance `delay` doesn't add alot of value as a standalone prop on `Transition`.

My original thought was to try and implement this _on top of_ the current current. e.g. wrap the components as they are so we don't need to make changes to the library here. I'd still like to try that approach before adding more code here, since these are already very delicate state machines.

Kovensky:
(off-topic but I want to spread Array.from)

Replace `[...new Array(10).keys()].map(() => ( <CSSFadeTransition/> )` with `Array.from({ length: 10 }, () => ( <CSSFadeTransition/> ))`; avoids using the array constructor which creates arrays with holes, and avoids having 2 intermediate arrays until you get the result of your map.

cjnaude:
This seems like a really useful addition.

Setsun:
Had some downtime and revived this PR. Going to fix the implementation to at least keep the conversation open about adding staggered transition support to `react-transition-group`.

Mikeysax:
I would love to see this added.

treardon17:
I would also like to see this added.

jquense:
Hey folks, My original comments still stand. I don't think that TransitionGroup or Transition are the correct places to add this sort of functionality. Animation delays are almost always the sort of thing you want to handle directly next to and with the animation itself. For instance with CSS transitions are used for in CSStransition and i'd expect that if you wanted to incorporate a delay used you the transition-delay property ideally, not a `setTimeout`. The other bit as well, is that doing this in JS in a animation agnostic way is probably not feasible. It's not enough to just do a `setTimeout` in JS, you'd need to coordinate with the animation itself , use requestAnimationFrame, deal with event loop timer drift, etc. All of which adds a lot of specific complexity to the base components that makes this one use-case easier but closes of others. 

I still think this is valuable feature and my original thought still stands, this could be implemented on top of this library, not integrated into it, using the currently available API. So someone could use the existing pieces here to build a StaggeredTransitionGroup

unruffledBeaver:
I've been working on a set of wrappers and stagger is part of my lib if anyone wants to check it out
https://github.com/unruffledBeaver/react-animation-components

Issue #94, opened 2017-07-07
Typo in v2.0.2 Doc on CSSTransition Section


Horizon-Blue:
**Do you want to request a *feature* or report a *bug*?**
bug

**What is the current behavior?**
On the section of [CSSTransition](https://reactcommunity.org/react-transition-group/#CSSTransition):

> And the coorresponding CSS for the <Fade> component:
> 
> .fade-enter {
>   opacity: 0.01;
> }
> 
> .fade-enter.fade-enter-active {
>   opacity: 1;
>   transition: opacity 500ms ease-in;
> }
> 
> .fade-leave {
>   opacity: 1;
> }
> 
> .fade-leave.fade-leave-active {
>   opacity: 0.01;
>   transition: opacity 300ms ease-in;
> }

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem via https://jsfiddle.net or similar (template: https://jsfiddle.net/1Lb2qk3v/).**
n/a

**What is the expected behavior?**
The last two css class names should be `fade-exit`, `fade-exit-active`, respectively, as described in the props section.

**Which versions, and which browser / OS are affected by this issue? Did this work in previous versions?**
v2.0.2

jquense:
Thanks for the issue, care to PR a fix?

paradoxxxzero:
This is wrong too:
```
timeout={{
 enter: 300,
 leave: 500,
}}
```

PR #96, opened 2017-07-07
Timeout example leave -> exit

paradoxxxzero:
I spent too much time figuring what goes wrong with the exit animations!
See #94


jquense:
Thanks!

PR #104, opened 2017-07-12
fix(Transition): it's onExited, not onExit

### LINKED ISSUE #103: TransitionGroup never unmounts children that left
Bug

Even after a child of TransitionGroup has left, it remains mounted forever

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem via https://jsfiddle.net or similar (template: https://jsfiddle.net/1Lb2qk3v/).**
```js
import React from 'react'
import range from 'lodash.range'

import Transition from 'react-transition-group/Transition'
import TransitionGroup from 'react-transition-group/TransitionGroup'

const duration = 300;

const defaultStyle = {
  transition: `opacity ${duration}ms ease-in-out`,
}

const transitionStyles = {
  entering: { opacity: 1 },
  entered:  { opacity: 1 },
  exiting: {opacity: 0},
  exited: {opacity: 0},
};

const Fade = ({ in: inProp, children }) => {
  return (
    <Transition appear enter exit in={inProp} timeout={duration}>
      {(state) => (
        <div style={{
          ...defaultStyle,
          ...transitionStyles[state]
        }}>
          {children}
        </div>
      )}
    </Transition>
  );
}

class TransitionGroupBug extends React.Component {
  static renderChild = number => <Fade key={number}>Child {number}</Fade>
  static zeroToOneHundred = range(0, 100).map(BasicTest.renderChild)
  static even = range(0, 100, 2).map(BasicTest.renderChild)
  static odd = range(1, 100, 2).map(BasicTest.renderChild)
  static prime = [
    2, 3, 5, 7, 11, 13, 17, 19, 23, 29, 31, 37, 39, 41, 43, 47, 53, 59, 61, 67, 71, 73, 79, 83, 89, 97,
  ].map(BasicTest.renderChild)

  state = {children: BasicTest.prime}

  render() {
    return (
      <div>
        <div>
          <button onClick={() => this.setState({children: BasicTest.zeroToOneHundred})}>0 - 100</button>
          <button onClick={() => this.setState({children: BasicTest.even})}>Evens</button>
          <button onClick={() => this.setState({children: BasicTest.odd})}>Odds</button>
          <button onClick={() => this.setState({children: BasicTest.prime})}>Primes</button>
        </div>
        <TransitionGroup enter exit {...this.state} />
      </div>
    )
  }
}
```

After the component mounts, click the `0 - 100` button, then the `Odds` button.

After transitions, `TransitionGroup` should have only 50 children.  Instead, it has 100 children, and the even numbers are merely faded out, leaving gaps between the odd numbers.

Yeah I know, adding `unmountOnExit` to `Transition` will make it **look** right.  *But that hides the fact that behind the scenes, an unbounded number of `<Transition in={false} />` components could accumulate!*

Chrome 59.0.3071.115 (Official Build) (64-bit)
Mac OS X

jedwards1211:
I think this fixes #103

### DIFF
--- a/src/Transition.js
+++ b/src/Transition.js
@@ -201,7 +201,7 @@ class Transition extends React.Component {
       });
       return;
     }
-    this.props.onExit(node);
+    this.props.onExited(node);
 
     this.safeSetState({ status: EXITING }, () => {
       this.props.onExiting(node);


jedwards1211:
nevermind, I was confused by the API.  Not sure what's causing #103

PR #108, opened 2017-07-12
docs(TransitionGroup): add Fade example

jedwards1211:
that shows passing through props to the `<Transition>` it renders

### DIFF
--- a/src/TransitionGroup.js
+++ b/src/TransitionGroup.js
@@ -66,8 +66,8 @@ const defaultProps = {
  *   exited:  { opacity: 0 },
  * };
  *
- * const Fade = ({ in: inProp, children, ...props }) => (
- *   <Transition in={inProp} timeout={duration} {...props}>
+ * const Fade = ({ children, ...props }) => (
+ *   <Transition timeout={duration} {...props}>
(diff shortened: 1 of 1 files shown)


src/TransitionGroup.js
```diff
@@ -49,7 +49,37 @@ const defaultProps = {
  * state machine for managing the mounting and unmounting of components over
  * time.
  *
- * Consider the example below using the `Fade` CSS transition from before.
+ * `<TransitionGroup>` clones its children with injected callbacks for the
+ * `<Transition>` component, so you should make sure to pass extra props through
+ * if your children are components that render a `<Transition>`.  For example:
+ *
+ * const duration = 300;
```
jquense: you need to surround the code blocks in backtick blocks, like on github. in fact it's github style markdown here.

jedwards1211: oh okay

silvenon:
We revamped the documentation and an example similar to this has been added in a CodeSandbox demo for `TransitionGroup`, but using `CSSTransition` instead. Because it's in a different format I'll have to close this PR, but thanks so much for the contribution!

PR #115, opened 2017-07-15
Clean up the README and migration guide a bit

taion:


taion:
Oh, boo. I guess renaming `migration.md` to `Migration.md` along with changing all the line wraps might have been a bit much for the ol' diff tracker to tell that it was a modification.

Still, I, uhh, cleaned things up a bit? Maybe?

PR #123, opened 2017-07-20
Fix Transition passes through all props to children in production build

### LINKED ISSUE #107: Transition passes through all props except component in production
bug

Because you are using [babel-plugin-transform-react-remove-prop-types](https://github.com/reactjs/react-transition-group/blob/master/.babelrc#L13), in production, [this line of `Transition`](https://github.com/reactjs/react-transition-group/blob/master/src/Transition.js#L270) does not delete anything from the props that it passes through:
```js
Object.keys(Transition.propTypes).forEach(key => delete childProps[key]);
```

`Transition` passes through the same set of props in production as it does in dev mode.

koba04:
Fix #107

### DIFF
--- a/src/Transition.js
+++ b/src/Transition.js
@@ -303,7 +303,21 @@ class Transition extends React.Component {
     }
 
     const {children, ...childProps} = this.props;
-    Object.keys(Transition.propTypes).forEach(key => delete childProps[key]);
+    // filter props for Transtition
+    delete childProps.in;
+    delete childProps.mountOnEnter;
+    delete childProps.unmountOnExit;
+    delete childProps.appear;
+    delete childProps.enter;
+    delete childProps.exit;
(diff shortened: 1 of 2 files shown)


src/Transition.js
```diff
@@ -303,7 +438,7 @@ class Transition extends React.Component {
     }
 
     const {children, ...childProps} = this.props;
-    Object.keys(Transition.propTypes).forEach(key => delete childProps[key]);
+    Object.keys(propTypes).forEach(key => delete childProps[key]);
```
jquense: lets do `delete props.foo` for each one manually instead, its a bit uglier but maintains the optimization

koba04:
@jquense Thanks! I've changed to delete the props manually and added a test for this.

koba04:
Thanks!

PR #124, opened 2017-07-20
support for multiple classes

adamborowski:
it is required when using react-jss with dynamic stylesheet - jss produces multiple class names then


jquense:
thanks!

PR #125, opened 2017-07-20
Update CSSTransition.js

adamborowski:
sorry but when running locally, I realized that classes variable has to be splitted before forEach :)


adamborowski:
can you publish patch version? otherwise I need to release scoped package on my own

PR #144, opened 2017-08-03
bug(CSSTransition): pass 'appearing' param from Transition to CSSTransition

alonbt:


jquense:
awesome thanks!

test/CSSTransition-test.js
```diff
@@ -95,19 +95,22 @@ describe('CSSTransition', () => {
...
           count++;
           expect(node.className).toEqual('custom custom-super-active');
+          expect(appearing).toBe(false);
         },
 
-        onEntered(node){
+        onEntered(node, appearing){
           expect(node.className).toEqual('');
           expect(count).toEqual(2);
+          expect(appearing).toBe(false);
```
jquense: could you also add a test confirming that it's `true` when appearing?

alonbt: I gave it a try but I'm not sure how to do it simple or how to do it at all. I tried to find an example for mounting a component but couldn't find on that file. I tried to do something like this but it didn't work:
```

    it.only('should apply pass isAppearing true', () => {

      let appearing = {}

      jest.resetModuleRegistry();
      jest.useFakeTimers();

      const container = document.createElement('div');
      const ReactDOM = require('react-dom');
      class Yolo extends React.Component {

        onEnter(node, isAppearing) {
          appearing.onEnter = isAppearing;
        }

        render() {
          return (
            <CSSTransition classNames="yolo" appear={true} enter={true} timeout={10} onEnter={this.onEnter}>
              <span>Spanning around</span>
            </CSSTransition>
          )
        }
      }

      ReactDOM.render(
        <Yolo/>,
        container,
      );

      jest.runAllTimers();

      expect(appearing.onEnter).toBe(true); // returns undefined === onEnter wasn't even called

    });
```

jquense: follow the same pattern as the Entering test, but where `instance` is 

```
instance = tsp(
        <CSSTransition
          timeout={10}
          appear={true}
          in={true}
          classNames="test"
        >
          <div/>
        </CSSTransition>
      )
      .render();
```

test/CSSTransition-test.js
```diff
@@ -112,6 +112,70 @@ describe('CSSTransition', () => {
         }
       });
     });
+
+    describe('isAppearing', () => {
```
alonbt: I refactored it to create 2 new tests - one for false and one for true

alonbt: it was also a bit confusing that this didn't work for the false part (because this pattern did work for the 'true' part):

```
it('should be false', done => {

        instance = tsp(
          <CSSTransition
            timeout={10}
            classNames="test"
            in={true}
          >
            <div/>
          </CSSTransition>
        )
          .render();


        instance.props({
          /* no in:true goes here..*/
         onEnter() ....
```

silvenon:
Fixed in df7adb45e3977ee93ddbb9eb087452bf78b872ae.

Issue #132, opened 2017-07-26
Incorrect exports for components in transpiled files


LKay:
There seems to be a *bug* in the compiled files.

**What is the current behavior?**
Currently compiled files, distributed on npm are not entirely correctly transpiled. The `Transition` component in `Transition.js` exporst component correctly using ES6 default export as:

```javascript
exports.default = Transition;
```

However the files `CSSTransition.js` and `TransitionGroup.js` also export that in commonjs styles as:

```javascript
exports.default = TransitionGroup;
module.exports = exports['default'];

//or

exports.default = CSSTransition;
module.exports = exports['default'];
```

This breaks ES6 imports in some configurations as they are "overriden" by cjs.

**What is the expected behavior?**

All files should be transpiled the same way. Preferably files should only contain ES6 exports as `index.js` exports everything as commonjs module anyway.

LKay:
This is realted to #109

ajgreenb:
I wonder why the transpiled exports would be different `TransitionGroup` and `CSSTransition`. In the source, they're defined the same:

* [`Transition`](https://github.com/reactjs/react-transition-group/blob/master/src/Transition.js#L499)
* [`TransitionGroup`](https://github.com/reactjs/react-transition-group/blob/master/src/TransitionGroup.js#L245)
* [`CSSTransition`](https://github.com/reactjs/react-transition-group/blob/master/src/CSSTransition.js#L272)

taion:
As noted, this is an issue with the TS defs. The exports are as intended here.

LKay:
@taion Exports in source files are correct. The problem lies in how Babel transpiles this code.

taion:
The top-level exports are exactly as they should be. They work just fine with Babel normally, and in fact follow the exact same pattern as in React: https://unpkg.com/react@15.6.1/lib/React.js.

There is no supposed "standard" for transpiled ES modules beyond whatever Babel happens to support. The only problem is with the TS defs.

LKay:
@taion I fixed TS definitions to reflect the actual transpiled code here https://github.com/DefinitelyTyped/DefinitelyTyped/pull/18622 but we end up having two different wat to import:
`import * as CSSTransition from "react-transition-group/CSSTransition"`
for `CSSTransition` due to presence of cjs export statement and
`import CSSTransition from "react-transition-group/Transition"`
in the file that's missing that statement. It's the way how Typescript compiler resolves modules...

taion:
[...]

PR #146, opened 2017-08-04
Change how ES6 modules are transpiled

### LINKED ISSUE #109: Standard ES6 imports fail
I'm using a standard npm install in an ES6 project that's compiled using typescript (ts-loader in webpack without babel), imports from docs fail for me:

```javascript
import TransitionGroup from 'react-transition-group/TransitionGroup';
```

While this works:
```javascript
import { TransitionGroup } from 'react-transition-group';
```

I might suggest updating the docs to use the 2nd syntax if it's less problematic, or listing both as options if it varies depending on the build environment.

LKay:
This seems to be preferred way of transpiling ES6 modules to enable common js imports as well https://babeljs.io/docs/plugins/transform-es2015-modules-commonjs/.

Currently use of `babel-add-export-modules` adds commonjs style `module.exports = exports["default"]` to transpiled files which breaks classic ES6 module imports (they are overwritten by this statement) in some environments such as Typescript. 

Fixes #109, #132

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -9,6 +9,7 @@
   "plugins": [
     "dev-expression",
     "transform-object-assign",
+    "transform-es2015-modules-commonjs",
     ["transform-react-remove-prop-types", { mode: "wrap" }]
   ],
 }
--- a/.babelrc
+++ b/.babelrc
@@ -8,7 +8,6 @@
   ],
   "plugins": [
     "dev-expression",
-    "add-module-exports",
     "transform-object-assign",
     ["transform-react-remove-prop-types", { mode: "wrap" }]
   ],
--- a/package.json
+++ b/package.json
@@ -59,7 +59,6 @@
     "babel-eslint": "^7.1.1",
     "babel-jest": "^20.0.3",
     "babel-loader": "^6.4.1",
-    "babel-plugin-add-module-exports": "^0.2.1",
     "babel-plugin-dev-expression": "^0.2.1",
     "babel-plugin-transform-object-assign": "^6.22.0",
     "babel-plugin-transform-react-remove-prop-types": "^0.4.5",
--- a/package.json
+++ b/package.json
@@ -60,6 +60,7 @@
     "babel-jest": "^20.0.3",
     "babel-loader": "^6.4.1",
     "babel-plugin-dev-expression": "^0.2.1",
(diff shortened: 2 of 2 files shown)


taion:
I don't think this is right – we don't want to break users of CJS modules, and this would have the effect of forcing CJS module users to have to add an extra `.default` when importing.

taion:
There's no such thing as ES module imports or exports. In practice all you have are various bundlers making their best guesses for supporting something that doesn't fully exist.

But anyway that doesn't even matter here – these are correctly-formed CJS modules.

LKay:
Well, Es6 modules and cjs modules are resolved slightly differently and thus adding `add-export-module` would break ES6 imports as it's monkey patching cjs `module.exports`. If you want to support cjs imports only then exporting default should be omitted and modules exported using `module.export =` statements. I'm totally fine with that but it should be unified everywhere so exports will be done one way or another.

taion:
There are no actual ES modules in the released code, period.

LKay:
Then how do you explain this in released code?:
```
exports.__esModule = true;
(...)
exports.default = CSSTransition;
module.exports = exports['default'];
```

jquense:
That's a common js module with a property "esmodule" that some bundles use as a convention like "pretend this is an es module". But they are all compiled to commonJS modules

LKay:
@jquense All but `Transition` component are exported that way. Is there a reason why this single file is exported differently without cjs export statement?

jquense:
> Is there a reason why this single file is exported differently without cjs export statement?

No, that is a bug! Transition should follow the same pattern

jquense:
Wait no, its the same. The output is due to how babel is compiling it. Transition has named exports so it can't export Transtion as `module.exports` and also export the various named constants.

taion:
Well, they should be importable in both ways. Just a bit tricky to get Babel to expose the right code.

LKay:
@jquense Maybe there is a way to force Babel to transpile source for `Transition` the same as the rest? If you look at the compiled code that's the only file that's missing `module.exports = exports['default']` statement and only has named export.

taion:
There is, but that's not the problem here. The issues TS users are reporting here are entirely resolved by https://github.com/DefinitelyTyped/DefinitelyTyped/pull/18637.

PR #187, opened 2017-09-06
fixed: appear bug on quickly updating components

jquense:


src/Transition.js
```diff
@@ -258,10 +258,19 @@ class Transition extends React.Component {
   }
 
   safeSetState(nextState, callback) {
+    // We need to track pending updates for instances where a cWRP fires quickly
```
taion: Is this a React 16 thing? How can a cWRP fire before cDM flushes?

jquense: its happening on v15 for me. I don't know. I'd imagine that's not behavior you want but it definitely seems to be happening

src/Transition.js
```diff
@@ -258,10 +258,19 @@ class Transition extends React.Component {
...
+    this.pendingState = nextState;
+
     // This shouldn't be necessary, but there are weird race conditions with
     // setState callbacks and unmounting in testing, so always make sure that
     // we can cancel any pending setState callbacks after we unmount.
-    this.setState(nextState, this.setNextCallback(callback));
+    callback = this.setNextCallback(callback)
+    this.setState(nextState, () => {
+      this.pendingState = null;
+      callback()
```
taion: semi

src/Transition.js
```diff
@@ -258,10 +258,19 @@ class Transition extends React.Component {
...
+    // We need to track pending updates for instances where a cWRP fires quickly
+    // after cDM and before the state flushes, which would double trigger a
+    // transition
+    this.pendingState = nextState;
+
     // This shouldn't be necessary, but there are weird race conditions with
     // setState callbacks and unmounting in testing, so always make sure that
     // we can cancel any pending setState callbacks after we unmount.
-    this.setState(nextState, this.setNextCallback(callback));
+    callback = this.setNextCallback(callback)
```
taion: semi

src/Transition.js
taion: weird

nmOobert:
I think I am hitting this bug. Is there a way I can install this change? Or what is an expected release of this?

STRML:
We hit the same. A v2.3.0 release would be very welcome so nobody else hits it.

jquense:
working on a release right now

PR #192, opened 2017-09-16
Wrong past tense?

TheoVitkovskiy:
I'm not sure about this one, but I think it should be "left" instead of leaved or is it what it was called in the first version?


jquense:
Yeah the "leaved" is intentional there, it's trying to illustrate the inconsistent of past tensing the old words if you try and do it symmetrically, resulting in a nonsense fake word "leaved"

PR #203, opened 2017-09-27
Allow React v16 peer dependency for v1 stable branch

lukeapage:


lukeapage:
thanks for the very quick response!

PR #211, opened 2017-09-29
Update to enzyme

jquense:


.travis.yml
```diff
@@ -5,6 +5,7 @@ node_js:
   - stable
 
 cache:
+  yarn: true
```
taion: i think you can just do `cache: yarn`

package.json
```diff
@@ -36,13 +37,16 @@
...
+      "./test/setup.js"
+    ],
     "roots": [
       "<rootDir>/test"
     ]
   },
   "peerDependencies": {
-    "react": "^15.0.0 || ^16.0.0",
-    "react-dom": "^15.0.0 || ^16.0.0"
+    "react": ">=15.0.0",
```
taion: `>= 15.3.0` because of the `PropTypes` thing

PR #212, opened 2017-09-29
Replace Transition

jquense:


stories/transitions/Fade.js
```diff
@@ -0,0 +1,41 @@
+import React from 'react';
+
+import CSSTransition from '../../src/CSSTransition';
+
+export const FADE_TIMEOUT = 3000;
+
+let styles = css`
```
taion: `const`!

stories/transitions/Fade.js
```diff
@@ -0,0 +1,41 @@
...
+  .exit {
+    opacity: 1;
+  }
+  .exit.exit-active {
+    opacity: 0.01;
+    transition: opacity 800ms ease-in;
+  }
+`;
+
+export default class Fade extends React.Component {
```
taion: Can we not use functional components here? Might be worth documenting why not.

LKay:
@jquense What is the status of this? 

I noticed this in the code and component is very small so far. Couldn't get it to work properly yet and I understand it's still in progress but maybe some snippet with example use case would be nice. 

Maybe you need some help with that if you could tell what is left and what are the issues? I created my own component for similar purpose some time ago https://github.com/LKay/react-transition-replace but haven't managed to properly transition height yet.

jquense:
its done and working but still not "release ready" quite yet, tho i'm using it in a few projects to dog food it a bit. There should be an example of how to use it in the storybook if you clone the repo locally and run those. Ih aven't really written documentation yet

LKay:
Yeah, I played with it a little bit. Looks like your `ReplaceTransition` is for slightly different purpose. It's more like swapping between two predefined `Transition`-like components depending on the `in` flag on wrapper. The naming confused me a little (maybe `SwapTransition` would be more correct).

levrik:
@jquense What happened to this feature? I'm searching for something like this for some time now already.

jquense:
@levrik it's in the library and released

levrik:
@jquense It's not documented and your last comment is that it's not "release ready" yet. Also the last commit that was made on the file has `[WIP]` in it. Not quite sure if I should use it in a production app.

jquense:
that's up to you to decide. I've not gotten around to writing docs, i'm not going to break semver with it tho so it's safe enough to use.

PR #220, opened 2017-10-04
Fix memleak in TransitionGroup

### LINKED ISSUE #219: Memory Leak in TransitionGroup's onExited
Bug
We're seeing memory leaks with React 16, apparently though the `onExited` handler here in `TransitionGroup`:

https://github.com/reactjs/react-transition-group/blob/8f971b853aef22f3c3b79a919856af5441062729/src/TransitionGroup.js#L164-L176

A screenshot of a heap dump from the leak:
screen shot 2017-10-04 at 1 50 02 pm

A second dump locally shows a reference to the previous child is being held in the closure created by `onExited`:

screen shot 2017-10-04 at 2 05 32 pm

I'm investigating the leak now, but I figured I'd put this issue in as a reference for a later PR, and to see if anyone else is experiencing this or has any leads.

STRML:
The closure created by the `onExited` handler prevented the previous
child from being cleaned up; this could cause chains tens of thousands
of elements long, consuming ever more memory as the application ran.

Fixes #219

### DIFF
--- a/src/TransitionGroup.js
+++ b/src/TransitionGroup.js
@@ -124,12 +124,8 @@ class TransitionGroup extends React.Component {
     // Initial children should all be entering, dependent on appear
     this.state = {
       children: getChildMapping(props.children, child => {
-        const onExited = (node) => {
-          this.handleExited(child.key, node, child.props.onExited);
-        }
-
         return cloneElement(child, {
-          onExited,
+          onExited: this.handleExited.bind(this, child),
           in: true,
           appear: this.getProp(child, 'appear'),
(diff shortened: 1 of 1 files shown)


taion:
I'm curious why this fixes things. Isn't this still closing over `child` via `bind`?

STRML:
The issue isn't closing over `child`, but keeping `prevChildMapping` in scope via that function - V8 doesn't appear to understand that it's never referenced and thus can be GC'd - so you get the long chains shown in the heapdump in #219.

PR #222, opened 2017-10-05
Upgrade peerDependencies of react react-dom

jasperjn:
Upgrade peerDependencies to accept React v16.0.0


jquense:
Hi there, thanks for the PR. The ranges here already accept v16 :)

PR #225, opened 2017-10-09
fix anchor link

romainwn:
The link was not redirecting to the anchor link like it was supposed to be. Now it is working


jquense:
thanks!

Issue #216, opened 2017-10-03
Transitions not applying when component mounts with React@16


igorrmotta:
**Do you want to request a *feature* or report a *bug*?**
Bug

**What is the current behavior?**
When trying to apply the transition on `componentDidMount`, the transition effect does not work.

Steps to reproduce:
1) create-react-app [project-name]
2) cd [project-name]
3) yarn add react-transition-group
4) Copy content from [here](https://gist.github.com/igorrmotta/36cbee9acc9f2d2360f28478ba1ff724) and paste on the `src/App.js`
5) Copy this code and past into `src/App.css`
```css
.fade {
  opacity: 0;
  transition: opacity 4s linear;
}
.fade.in {
  opacity: 1;
}
```

**What is the expected behavior?**
The transition effect should be applied when the `componentDidMount` is fired. However, it doesn't happen. 
If you comment the [line 46](https://gist.github.com/igorrmotta/36cbee9acc9f2d2360f28478ba1ff724#file-bug-react16-with-reacttransitiongroup-jsx-L46) and uncomment the [setTimeout(() on line 48, 49, 50, 51](https://gist.github.com/igorrmotta/36cbee9acc9f2d2360f28478ba1ff724#file-bug-react16-with-reacttransitiongroup-jsx-L48). **It will fix the bug**.

**Which versions, and which browser / OS are affected by this issue? Did this work in previous versions?**
This bug only reproduces with the latest version of React (16).

*something tells me that it is related to the new [react fiber architecture](https://github.com/acdlite/react-fiber-architecture)*

tobias-zucali:
I think I have the same problem… This is just the samples for `Transition` & `TransitionGroup`from the docu page merged together.

**What is the expected behavior?**
Items are faded in as they are added.
The function inside `Transitionin` is called with state `exited`, `entering` and `entered`, if I place a debugger there the animation works fine. Without the debugger there is no fade effect.
It seems that React 16 optimised the rendering so well that there is no redraw cycle between `exited` and `entering`, which destroys our animation effect.

**Which versions, and which browser / OS are affected by this issue?**
react: ^16.0.0
react-dom: ^16.0.0
react-transition-group: ^2.2.1


``` js
import React from 'react';
import TransitionGroup from 'react-transition-group/TransitionGroup';
import Transition from 'react-transition-group/Transition';

const duration = 300;

const defaultStyle = {
  transition: `opacity ${duration}ms ease-in-out`,
  opacity: 0,
}

[...]

PR #227, opened 2017-10-15
Workaround issue #216

yannisgu:
Extracted the initial transition from EXITED to ENTERING to setTimeout, so that a paint will be done between. Else it is not possible to do CSS transitions, as the browser will not interpret changes to CSS properties as change.

### DIFF
--- a/src/Transition.js
+++ b/src/Transition.js
@@ -137,7 +137,9 @@ class Transition extends React.Component {
   }
 
   componentDidMount() {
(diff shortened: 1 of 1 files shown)


jquense:
Hey there thanks for the PR, we are intentionally keeping the Transition api low level and not specific to css transitions. You can look back through previous issues on this and their various fixes, or poke through the source of CSSTransition itself to see how it deals with this. Thanks again!

PR #228, opened 2017-10-17
TransitionGroup: moving handleExited to the prototype

dimitarnestorov:
Since everywhere the method is being used is bound to `this`


jquense:
thanks!

PR #232, opened 2017-10-18
Fix CSSTransition codesandbox example

alaycock:
The existing example codesandbox for the CSSTransition component doesn't properly hide the `div`. The `div` fades out, then reappears without a transition, as soon as it's done fading out. I updated the example to include the `unmountOnExit` prop, so it properly shows and hides the `div`.

Existing:
https://codesandbox.io/embed/yv645oq21x

New:
https://codesandbox.io/embed/kw8z6pp9zo


jquense:
Thanks!

PR #236, opened 2017-10-24
Add es6 module build

LKay:
Constants from `Transition.js` were moved to `constants.js` file to fix incorrect default export for CommonJS output (`babel-add-module-exports` was ignoring that).


LKay:
Travis fails for unknown reason... All tests seem to be passing when running `npmt test` locally:

```
PASS  test/Transition-test.js
  Transition
    ✓ should not transition on mount (18ms)
    ✓ should transition on mount with `appear` (5ms)
    ✓ should pass filtered props to children (4ms)
    ✓ should allow addEndListener instead of timeouts (9ms)
    ✓ should fallback to timeous with addEndListener  (5ms)
    entering
      ✓ should fire callbacks (16ms)
      ✓ should move to each transition state (15ms)
    exiting
      ✓ should fire callbacks (15ms)
      ✓ should move to each transition state (16ms)
    mountOnEnter
      ✓ should mount when entering (4ms)
      ✓ should stay mounted after exiting (26ms)
    unmountOnExit
      ✓ should mount when entering (3ms)
      ✓ should unmount after exiting (19ms)

 PASS  test/TransitionGroup-test.js
  TransitionGroup
    ✓ should allow null components (127ms)
    ✓ should allow callback refs (5ms)
    ✓ should work with no children (1ms)
    ✓ should handle transitioning correctly (12ms)
    ✓ should not throw when enter callback is called and is now leaving (4ms)
    ✓ should not throw when leave callback is called and is now entering (3ms)

 PASS  test/CSSTransitionGroup-test.js
  CSSTransitionGroup
    ✓ should clean-up silently after the timeout elapses (25ms)
    ✓ should keep both sets of DOM nodes around (14ms)
    ✓ should switch transitionLeave from false to true (17ms)
    ✓ should work with a null child (15ms)
    ✓ should transition from one to null (11ms)
    ✓ should transition from false to one (10ms)
    ✓ should clear transition timeouts when unmounted (9ms)
    ✓ should handle unmounted elements properly (9ms)
    ✓ should work with custom component wrapper cloning children (8ms)

 PASS  test/CSSTransition-test.js
  CSSTransition
    ✓ should flush new props to the DOM before initiating a transition (6ms)
    entering
      ✓ should apply classes at each transition state (16ms)
      ✓ should apply custom classNames names (19ms)
    exiting
      ✓ should apply classes at each transition state (19ms)
      ✓ should apply custom classNames names (17ms)

 PASS  test/ChildMapping-test.js
  ChildMapping
    ✓ should support getChildMapping (2ms)
    ✓ should support mergeChildMappings for adding keys (1ms)
    ✓ should support mergeChildMappings for removing keys (1ms)
    ✓ should support mergeChildMappings for adding and removing
    ✓ should reconcile overlapping insertions and deletions (1ms)
    ✓ should support mergeChildMappings with undefined input (1ms)

Test Suites: 5 passed, 5 total
Tests:       39 passed, 39 total
Snapshots:   0 total
Time:        3.774s
Ran all test suites.
```

.babelrc
```diff
@@ -1,15 +1,70 @@
...
+        "stage-2"
+      ],
+      "plugins": [
+        "dev-expression",
+        "add-module-exports",
+        "transform-object-assign",
+        ["transform-react-remove-prop-types", { "mode": "wrap" }]
+      ]
+    },
+    "testing": {
```
taion: key should be "test"

.babelrc
```diff
@@ -1,15 +1,70 @@
...
-  ],
-  "plugins": [
-    "dev-expression",
-    "add-module-exports",
-    "transform-object-assign",
-    ["transform-react-remove-prop-types", { "mode": "wrap" }]
-  ]
+  "env" : {
+    "development": {
+      "presets": [
```
taion: if we don't want to use a babel 7, `.babelrc.js`, then we should just make this a preset like https://github.com/relay-tools/react-router-relay/blob/v0.14.0/tools/latestPreset.js to avoid duplicating the code a bunch

.babelrc
```diff
@@ -1,15 +1,70 @@
...
+        "stage-2"
+      ],
+      "plugins": [
+        "dev-expression",
+        "add-module-exports",
+        "transform-object-assign",
+        ["transform-react-remove-prop-types", { "mode": "wrap" }]
+      ]
+    },
+    "es6": {
```
taion: key should be "es"

package.json
```diff
@@ -2,12 +2,15 @@
...
+  "main": "index.js",
+  "module": "es6/index.js",
   "scripts": {
     "test": "npm run lint && npm run testonly",
     "testonly": "jest --verbose",
     "tdd": "jest --watch",
-    "build": "rimraf lib && babel src --out-dir lib && npm run build:dist && cp README.md LICENSE ./lib",
+    "build": "npm run build:lib && npm run build:es6 && npm run build:dist && cp README.md LICENSE ./lib",
+    "build:lib": "rimraf lib && babel src --out-dir lib",
+    "build:es6": "rimraf lib/es6 && BABEL_ENV=es6 babel ./src --out-dir ./lib/es6",
```
taion: `build:es`, and output to `lib/es/` instead of `lib/es6/`

src/index.js
```diff
@@ -1,12 +1,4 @@
...
-import Transition from './Transition';
-
-
-module.exports = {
-  Transition,
-  TransitionGroup,
-  ReplaceTransition,
-  CSSTransition,
-};
+export { default as CSSTransition } from './CSSTransition';
```
taion: This changes the semantics for existing users, which would be a breaking change.

src/Transition.js
```diff
@@ -3,12 +3,7 @@ import React from 'react';
 import ReactDOM from 'react-dom';
 
 import { timeoutsShape } from './utils/PropTypes';
-
-export const UNMOUNTED = 'unmounted';
-export const EXITED = 'exited';
-export const ENTERING = 'entering';
-export const ENTERED = 'entered';
-export const EXITING = 'exiting';
+import { UNMOUNTED, EXITED, ENTERING, ENTERED, EXITING } from './constants';
```
taion: This is also a breaking change.

LKay: This is only used (imported) in tests and not documented anywhere so I considered these exports part of private API. These however break `babel-add-module-exports` plugin behaviour as it not adding `module.exports = export['default']` at the end of file when there are other named exports present.

src/index.js
LKay: How does that change the result of below?:
```js
import { Transition, CSSTransition, TransitionGroup } from 'react-transition-group';
// or cjs
const { Transition, CSSTransition, TransitionGroup } = require('react-transition-group');
```
Transpiled code works exactly the same as before.

src/Transition.js
jquense: they are public, if not well documented.

test/TransitionGroup-test.js
```diff
@@ -13,7 +13,7 @@ describe('TransitionGroup', () => {
   beforeEach(() => {
     React = require('react');
     ReactDOM = require('react-dom');
-    Transition = require('../src/Transition').default;
+    Transition = require('../src/Transition');
```
LKay: This is the only breaking change (for cjs `require` users) I can see here, so they won't have to type `.default` for `Transition` anymore. That how it should have worked from the beginning, though.

src/Transition.js
LKay: I see. They seem to be in conflict with static properties on `Transition` class and that is probably the reason why `babel-add-module-exports` fails to add the statement.

jquense: the failure there is due to how the transpilation works with a mix of named and default exports. essentially it doesn't want to assign the exports to Transition implicitly. We can "fix" this maybe by manually doing it

LKay: This is not recommended way as described here: https://github.com/59naga/babel-plugin-add-module-exports/issues/27 and in the referenced issues. As long as the file contains multiple exports this would never work out of the box.

jquense: right That's why i'm suggesting doing it manually. It's not that it won't work its that it would be destructive for the plugin to do automatically since it might bulldoze over existing properties on the default export

LKay: Ok, I added a hack to manually insert the cjs default export statement to `Transition`.

silvenon:
There was another much older PR for ES6 modules, #77, which I just merged. A lot of effort went into both of these PRs, so I'm sad to have to close this one. Thanks for contributing! ❤️

PR #243, opened 2017-11-01
render only children with component=null

### LINKED ISSUE #242: React16 - TransitionGroup without a wrapping div
Bug

I have to pass in a custom component or custom styles that are child-aware

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem via https://jsfiddle.net or similar (template: https://jsfiddle.net/m4mb9Lg3/4/).**

1. Render a component like: `<div style={{width: '100%}}></div>`
2. Now wrap it in a TransitionGroup
3. Notice the width is borked & you have to apply `width: 100%` to the TransitionGroup

I can add a transition by simply wrapping a component in TransitionGroup. Since React 16 allows you to return an array of elements, there's no need for a div. If backwards compatibility is desired, we could use special handling if `component=null`. Happy to PR if this has a chance of getting merged.

All versions

mattkrick:
Fixes #242 

Usage:
```
<TransitionGroup component={null}>
  <div style={{width: '100%'}}></div>
</TransitionGroup>
```

I'd love to see this become the default behavior. TransitionGroup is a container, and probably shouldn't modify the DOM. With React 16, now we don't have to.

### DIFF
--- a/src/TransitionGroup.js
+++ b/src/TransitionGroup.js
@@ -179,15 +179,18 @@ class TransitionGroup extends React.Component {
 
   render() {
     const { component: Component, childFactory, ...props } = this.props;
-    const { children } = this.state;
+    const children = values(this.state.children).map(childFactory);
 
     delete props.appear;
     delete props.enter;
     delete props.exit;
 
+    if (Component === null) {
+      return children;
+    }
     return (
       <Component {...props}>
-        {values(children).map(childFactory)}
+        {children}
       </Component>
     );
   }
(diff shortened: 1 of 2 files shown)


taion:
Changing the default isn't really possible at the moment – it'd be a breaking change in both behavior and in dropping React 15 support.

jquense:
this doesn't change the default tho?

taion:
In regard to
> I'd love to see this become the default behavior.

mattkrick:
Correct, no change to default behavior! semver minor & doesn't break the current `Component` test.

jquense:
can you document this behavior in the right jsdoc comment and then we should be gtg

mattkrick:
ah good call. documented!

jquense:
looks good, just need to get CI to go green (looks like lint issues)

jquense:
Thanks!

Billy-:
When will this be released? Thanks

jquense:
when i get around to it

PR #258, opened 2017-11-26
Update documentation

newyork-anthonyng:


taion:
The docs are referring to the exported constants in https://github.com/reactjs/react-transition-group/blob/df4f6e48addacd38886af292e6a0ecf53772a2fa/src/Transition.js#L8-L11

newyork-anthonyng:
@taion I made the updates; thanks for catching that.

taion:
Thanks!

PR #268, opened 2017-12-21
remove `classnames` from dependencies

1pete:


taion:
We don’t use this at all? Cool.

Issue #261, opened 2017-12-01
How do you remain in an entered state using CSSTransition?


studious:
Hi all.

I am using CSSTransition so that I can use keyframe animations for enter and exit. The challenge I have is that when the transitioning classnames disappear I am back at the start. I have tried using the onExited and onEntered callback to add the destination classname but they fire just 'after' the classnames have been removed which introduces a micro flicker of the original style and my new destination style. Is there some way to handle this?

```
.draw--closed {
  display: block;
  background: url('./assets/TilesDraw-background.png');
  margin-top: -10px;
  margin-left: -100vw;
  height: 0;
  width: 100vw;
}

.draw--opened {
  display: block;
  background: url('./assets/TilesDraw-background.png');
  margin-top: -10px;
  margin-left: 0vw;
  height: 429px;
  width: 100vw;
}

.draw-enter {
  height: 0;
  margin-left: -100vw;
}
.draw-enter.draw-enter-active {
  animation-name: slideIn;
  animation-duration: 2s;
}
.draw-exit {
  height: 429px;
  margin-left: 0vw;
}
.draw-exit.draw-exit-active {
  animation-name: slideOut;
  animation-duration: 2s;
}
@keyframes slideIn {
  0% {
    height: 0px;
    margin-left: -100vw;
  }
  50% {
    height: 429px;
    margin-left: -100vw;
  }
  100% {
    height: 429px;
    margin-left: 0vw;
  }
}
@keyframes slideOut {
  0% {
    margin-left: 0vw;
    height: 429px;
  }
  50% {
    margin-left: -100vw;
    height: 429px;
  }
  100% {
    margin-left: -100vw;
    height: 0px;
  }
}
```

ryanwang520:
Yep, I also think this is a problem, and I wonder why not add an `-entered` classname when the animation ends.

Mikhail-Fiadosenka:
Maybe [CSS property](https://developer.mozilla.org/en-US/docs/Web/CSS/animation-fill-mode) `animation-fill-mode` can help to achieve the desired effect?

studious:
Did you think I could get by without changing the classname in the final callback?

I'm finding I still need to use `onEntered` and `onExited` to alter the classname but using `animation-fill-mode` seems to stop the flash of original styling I was experiencing. I haven't tested in all browsers yet but it looks promising.

So at this stage it might be nice to have an `-entered` classname on the end to get around my need to two callbacks that alter the classname themselves.

[...]

PR #269, opened 2017-12-22
Add -done class names

GMchris:
New "Done" states as discussed [here](https://github.com/reactjs/react-transition-group/issues/261).

Before this change there was no way to persist the styles that have been transitioned to, without fiddling with the callbacks. Now users can define `#{transition-name}-done` classes or custom ones to denote the transition as being done and keep the element in the new state or set other styles for it. 

Here's a small [Sass mixin](https://gist.github.com/GMchris/fae64231d2a2ae07fb294c54aafe7189) that generates css compliant with this change.

### DIFF
--- a/src/CSSTransition.js
+++ b/src/CSSTransition.js
@@ -27,8 +27,10 @@ const propTypes = {
    *  appearActive: 'my-active-appear',
    *  enter: 'my-enter',
    *  enterActive: 'my-active-enter',
+   *  enterDone: 'my-done-enter,
    *  exit: 'my-exit',
    *  exitActive: 'my-active-exit',
+   *  exitDone: 'my-done-exit,
    * }}
    * ```
    *
@@ -37,8 +39,10 @@ const propTypes = {
    *  appearActive?: string,
    *  enter?: string,
    *  enterActive?: string,
+   *  enterDone?: string,
    *  exit?: string,
    *  exitActive?: string,
+   *  exitDone?: string,
    * }}
    */
   classNames: classNamesShape,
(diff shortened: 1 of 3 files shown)


taion:
Trivial question, but `enterDone`/`exitDone` v `entered`/`exited`?

jquense:
Done is more in keeping with the existing verbage of the CSSTransition, e.g. enter enter-active enter-done, vs enter, enter-active, entered

we may want to change it tho to match Transition at somepoint

GMchris:
Hey, just wanted to check if there's something you think needs to be changed to merge these changes. Have you decided on 'done' vs 'entered' for instance?

jquense:
looks good can you update the doc comments to explain the new classes? thanks

jquense:
thanks!

guidobouman:
@jquense When is this getting released? I'm currently getting bitten by this.

We can only use class names because we're using `CSSTransition` in combination with Glamorous. The class names that I pass along to `CSSTransition` have a lower specificity than the default styling which also uses classes...

RanzQ:
Anyone knows a workaround to set a base className when there's no transition in progress, until this is released? Only way I found is to use a wrapper element just for the transition.

studious:
@RanzQ I use a default class to start with. I then set my own done classname using onEntered and onExited webhooks.

```
const Example = ({isSelected, children}) => {
  const onEntered = node => {
    node.className = 'example-selected';
  };
  const onExited = node => {
    node.className = 'example';
  };

  return (
    <CSSTransition
      in={isSelected}
      appear={true}
      timeout={1000}
      classNames="example"
      onEntered={onEntered}
      onExited={onExited}
      >{children}</CSSTransition>
  )
};
```

oliverwoodings:
Is there any news on when this is going to be released?

Issue #208, opened 2017-09-28
Uncaught TypeError: Cannot read property 'classList' of null


Mikeysax:
**Do you want to request a *feature* or report a *bug*?**
Bug
**What is the current behavior?**
When I try to toggle the Signout modal, I get this error:
```
removeClass.js:4 Uncaught TypeError: Cannot read property 'classList' of null
    at removeClass (removeClass.js:4)
    at CSSTransition.js:46
    at Array.forEach (<anonymous>)
    at removeClass (CSSTransition.js:45)
    at CSSTransition.removeClasses (CSSTransition.js:278)
    at Object.CSSTransition._this.onEnter (CSSTransition.js:214)
    at Transition.performEnter (Transition.js:242)
    at Transition.updateStatus (Transition.js:213)
    at Transition.componentDidUpdate (Transition.js:179)
    at commitLifeCycles (react-dom.development.js:11517)
removeClass @ removeClass.js:4
(anonymous) @ CSSTransition.js:46
removeClass @ CSSTransition.js:45
removeClasses @ CSSTransition.js:278
CSSTransition._this.onEnter @ CSSTransition.js:214
performEnter @ Transition.js:242
updateStatus @ Transition.js:213
componentDidUpdate @ Transition.js:179
commitLifeCycles @ react-dom.development.js:11517
commitAllLifeCycles @ react-dom.development.js:12294
callCallback @ react-dom.development.js:1299
invokeGuardedCallbackDev @ react-dom.development.js:1338
invokeGuardedCallback @ react-dom.development.js:1195
commitAllWork @ react-dom.development.js:12415
workLoop @ react-dom.development.js:12687
callCallback @ react-dom.development.js:1299
invokeGuardedCallbackDev @ react-dom.development.js:1338
invokeGuardedCallback @ react-dom.development.js:1195
performWork @ react-dom.development.js:12800
batchedUpdates @ react-dom.development.js:13244
performFiberBatchedUpdates @ react-dom.development.js:1646
stackBatchedUpdates @ react-dom.development.js:1637
batchedUpdates @ react-dom.development.js:1651
batchedUpdatesWithControlledComponents @ react-dom.development.js:1664
dispatchEvent @ react-dom.development.js:1874
react-dom.development.js:8305 The above error occurred in the <Transition> component:
    in Transition
    in CSSTransition (created by App)
    in div (created by App)
    in div (created by App)
    in App (created by Connect(App))
    in Connect(App) (created by RouterContext)
[...]

PR #270, opened 2018-01-05
Fix "Uncaught TypeError: Cannot read property 'classList' of null"

orfjackal:
This is a fix for issue #208 

This fix has also been published at https://www.npmjs.com/package/@luontola/react-transition-group as 2.3.0-patch.3

### DIFF
--- a/src/CSSTransition.js
+++ b/src/CSSTransition.js
@@ -7,8 +7,8 @@ import React from 'react';
 import Transition from './Transition';
 import { classNamesShape } from './utils/PropTypes';
 
-const addClass = (node, classes) => classes && classes.split(' ').forEach(c => addOneClass(node, c));
-const removeClass = (node, classes) => classes && classes.split(' ').forEach(c => removeOneClass(node, c));
+const addClass = (node, classes) => node && classes && classes.split(' ').forEach(c => addOneClass(node, c));
+const removeClass = (node, classes) => node && classes && classes.split(' ').forEach(c => removeOneClass(node, c));
 
 const propTypes = {
   ...Transition.propTypes,
@@ -190,7 +190,7 @@ class CSSTransition extends React.Component {
     // This is for to force a repaint,
     // which is necessary in order to transition styles when adding a class name.
     /* eslint-disable no-unused-expressions */
-    node.scrollTop;
+    node && node.scrollTop;
     /* eslint-enable no-unused-expressions */
     addClass(node, className);
   }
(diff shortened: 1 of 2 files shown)


src/Transition.js
```diff
@@ -187,6 +187,9 @@ class Transition extends React.Component {
       // nextStatus will always be ENTERING or EXITING.
       this.cancelNextCallback();
       const node = ReactDOM.findDOMNode(this);
+      if (!node) {
+        return;
+      }
```
jquense: I'm not sure this is enough to solve the problem. When is the node not available? this will leave the component in a bad state if it had to do anything else, does this situation mean the component is already unmounted?

orfjackal: Our use case is that we have complex forms where form fields are dynamically added and removed based on the values of other form fields. They are rendered like this:

```jsx
<TransitionGroup>
  {fields.map(field =>
    <CSSTransition {...transitionProps}>
      {field}
    </CSSTransition>)}
</TransitionGroup>
```

The rendering of some of those fields also has many conditionals, including rendering the field as null:

```jsx
const SomeField = ({someProp}) => {
  if (someProp === 'foo') {
    return <div>...</div>;
  }
  if (someProp === 'bar') {
    return <div>...</div>;
  }
  return null;
};
```

As a result, sometimes CSSTransition's child component will have no DOM node. This used to work fine with react-transition-group v1.2.0, but with v2.2.1 it causes the crash mentioned in issue #208.

jquense: ok, so much concern here is that this effectively skips an animation, but doesn't clean up the state. So now you'd have the Transition Component props saying it's entered, but its internal state still saying it's exited.

I think this need to handled in `CSSTranstion`, which uses the DOM node as opposed to here which is mostly DOM agnostic. Said another way, i think transitioning a component where the DOM node is null is a valid use case, it should be handled up one level in CSSTransition where it's not a valid case

orfjackal: Yes, handling it inside CSSTransition sounds best. My first attempt was to add null checks to addClass and removeClass in CSSTransition.js, but those two places were not enough, so I backed out before going too deep into that rabbit hole.

jquense: it should be eneough to return early from all of the `onFoo` hooks there

orfjackal:
Here is an updated PR, which only modifies CSSTransition and also adds a test for the bug.

jquense:
thanks!

jpfiorilla:
solved the issue for me, thanks a bunch

mreishus:
is there going to be a new release of react-transition-group that includes this fix?

PR #273, opened 2018-01-10
Update README.md

trevorwhealy:
Different was not as clear to me -- these are not backward compatible. 
Technically semantic versioning implies this, but it's better to be explicit.


trevorwhealy:
@jquense

jquense:
thanks!

PR #281, opened 2018-01-25
Fixed continously rebound handleExited in TransitionGroup

devfive:
Follow up on #219

### DIFF
--- a/src/TransitionGroup.js
+++ b/src/TransitionGroup.js
@@ -89,7 +89,7 @@ class TransitionGroup extends React.Component {
     this.state = {
       children: getChildMapping(props.children, child => {
         return cloneElement(child, {
-          onExited: this.handleExited.bind(this, child),
+          onExited: this.handleExited.bind(this, child.key, child.props.onExited),
           in: true,
           appear: this.getProp(child, 'appear'),
           enter: this.getProp(child, 'enter'),
@@ -136,7 +136,7 @@ class TransitionGroup extends React.Component {
       if (hasNext && (!hasPrev || isLeaving)) {
         // console.log('entering', key)
         children[key] = cloneElement(child, {
-          onExited: this.handleExited.bind(this, child),
+          onExited: this.handleExited.bind(this, key, child.props.onExited),
           in: true,
           exit: this.getProp(child, 'exit', nextProps),
           enter: this.getProp(child, 'enter', nextProps),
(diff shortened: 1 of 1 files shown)


STRML:
So this is purely a performance boost, yes? Do you have any profiles that show the extent?

taion:
This won’t work if the child onExited changes, though, will it?

taion:
Last I heard, arrow functions were much faster than bind, so perhaps we can get a partial speedup without the semantics compromise by using an arrow function.

jquense:
i think this used to be an arrow function. the main problem is not the `bind` perf but the closure created by the arrow function (if i'm remembering correctly)

STRML:
> Last I heard, arrow functions were much faster than bind,

That changed around Chrome 54. It's been a while, bind is now properly optimized. http://benediktmeurer.de/2015/12/25/a-new-approach-to-function-prototype-bind/

jquense:
I'm definately gald to have input and contributions on the perf front, however i don't think this sort of optimization is worth it without contextualized numbers, especially since it's changing semantics

STRML:
I'd be very very surprised if a single bind had any measurable performance
impact compared to the rest of the render.

Xizario:
@jquense, This one is not just a semantics change, it reverses a breaking change that was made in https://github.com/reactjs/react-transition-group/issues/219 
After 219 the `child.props.onExited(node)`; changes `this` of our handler that we pass as `onExited`

@STRML, This pull request is from 25 Jan, why it was not merged already? All of the sudden v2.3.0 got official and breaks our existing code. This PR would have avoided that, but was never merged.

jquense:
It wasn't merged because it changed semantics and then not updated.

If something else broke please open a separate issue

Xizario:
@jquense, if this is the case, why https://github.com/reactjs/react-transition-group/pull/220 was merged in the first place? 

Your arguments are invalid, because 2.3.0 was not released back when this PR was created. Where is the point of opening an issue? I already see that you don't care for look them, nor the PRs.

jquense:
Your point seems to be we should introduce new bugs intentionally because an other PR also introduced a bug? I'm not sure sure waht you are even angry about these PRs are about fixing a memory leak and performance. #220 doesn't change what `this` is in your handler so your angry is misplaced, if you want something addressed be productive about it, don't chastise open source volunteers

PR #282, opened 2018-01-27
Fix TransitionGroup error for v1 as well

### LINKED ISSUE #12: Broken animation finishing after unmounting

jony89:
fixes #12 using #15 for v1 as well.

Not all of us have moved to v2 (especially those of us who are still using `material-ui`  < v1-beta),
Yet, this is critical bug fix .

@jquense hopefully a version with that fix will be released soon ? :)

### DIFF
--- a/src/TransitionGroup.js
+++ b/src/TransitionGroup.js
@@ -99,7 +99,7 @@ class TransitionGroup extends React.Component {
   };
 
   _handleDoneAppearing = (key, component) => {
-    if (component.componentDidAppear) {
+    if (component && component.componentDidAppear) {
       component.componentDidAppear();
     }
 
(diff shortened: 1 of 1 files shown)


jquense:
thanks!

jony89:
@jquense could u publish a new version for this ? (v1.2.2)

Issue #159, opened 2017-08-14
Not enough time between 'exited' and 'entering' for DOM to perform enter transition


jalooc:
**Do you want to request a *feature* or report a *bug*?**
Bug (or there's sth wrong with my code)

**What is the current behavior?**
During entering, the `exited` state lasts for minimal amount of time, but sometimes it appears to be too little, such that the browser doesn't render the style associated to that state and the enter transition doesn't work.

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem via https://jsfiddle.net or similar (template: https://jsfiddle.net/m4mb9Lg3/4/).**
https://jsfiddle.net/m4mb9Lg3/25/
Note, that exit transition works properly, whilst entering is not transitioned. In console you can see however, that the `state` prop is passed correctly and goes through it's whole cycle.

The other day, doing _enter_ animations in pure React stateful component (no `react-transition-group`), something similar occurred:

I would have an `enteredActive` state, initially set to `false`. In `componentDidMount`, inside `requestAniationFrame` callback I would set it to `true`. This would trigger rendered styles change and thus animation. That's fine except it's not, because oftentimes (not always) this wouldn't work. What fixed it permanently, was to put the `rAF` inside another `rAF`, like this:
```javascript
componentDidMount() {
  requestAnimationFrame(() =>
    requestAnimationFrame(
      () => this.setState({ enteredActive: true })
    )
  )
}
```
My suspicion is that it has something to do with React internals.

So to the point: maybe the double-raf-technique inside `Transition` component would help? Or yet, there's sth wrong with my codepen?

**Which versions, and which browser / OS are affected by this issue? Did this work in previous versions?**
All of them.

orballo:
I have this same problem. I found an old issue about this same topic but in the CSSTransition component. I can't locate it right now but they solved it forcing a reflow using `node.scrollTop`.  Here is the [code](https://github.com/reactjs/react-transition-group/blob/master/src/CSSTransition.js#L242-L249):
```javascript
reflowAndAddClass(node, className) {
    // This is for to force a repaint,
    // which is necessary in order to transition styles when adding a class name.
    /* eslint-disable no-unused-expressions */
    node.scrollTop;
    /* eslint-enable no-unused-expressions */
    addClass(node, className);
}
```
So, we used this same solution in the `onEnter` callback and works fine:
```javascript
<Transition
  in={isOpen}
  timeout={350}
  mountOnEnter
  unmountOnExit
[...]

PR #304, opened 2018-03-16
Split docs into pages, update examples and other details

silvenon:
Hi! This is a work in progress, but I'm opening it now to start a discussion. Basically I would like to help a lot with the documentation. My idea would be to first split the documentation into separate pages so it's easier to navigate. Afterwards I would like to improve some examples and recommend using `CSSTransition` instead of `Transition` for animating CSS properties because of the reflow hack, I wrote a [blog post about it](https://blog.silvenon.com/the-nuances-of-react-transition-group-9214b2302138) because I noticed quite a few of issues like [this one](https://github.com/reactjs/react-transition-group/issues/159#issuecomment-322761171).

Here's a preview of what I got so far.

rtg-docs-split-pages

I also wanted to update the development environment:

- I renamed `.eslintrc` files to `.eslintrc.yml` to get syntax highlighting (they were already written in YAML)
- I fixed some ESLint errors and added some more based on conventions of the existing code
- more changes to come to make development and contribution easier

What do you think?


jquense:
Hey this is awesome! Thanks for taking the time to work on this. I think your plan sounds great

silvenon:
Great! Would you be ok if I potentially switched Bootstrap with [Semantic UI](https://react.semantic-ui.com)? The reason I'm asking is primarily because we have to add tons of classes and other attributes which could otherwise be abstracted away, for example look at the code required for one toggle button for mobile navigation:

```jsx
<button className='navbar-toggler' type='button' data-toggle='collapse' data-target='#navbarSupportedContent' aria-controls='navbarSupportedContent' aria-expanded='false' aria-label='Toggle navigation'>
  <span className='navbar-toggler-icon' />
</button>
```
Also, we need to add global jQuery for any JS functionality, like the toggle button above.

jquense:
I'm not familiar with semantic ui other than I know it's a ui library, I am a maintainer if react-bootstrap tho so I'm a lot more familiar with it.

Are we using RB here it did I just slap on normal bootstrap? RB should equally abstract a lot a way, including many a11y bits

silvenon:
We currently have normal Bootstrap. I initially went for your react-bootstrap, but it uses Bootstrap 3, understandably, because v4 isn't stable yet. However, Bootstrap 3 looks… old. 😅😅😅 Would you be ok with [reactstrap](https://reactstrap.github.io/) which implements Bootstrap 4? React Transition Group's documentation is a really simple site, so I don't think we can go wrong here, I just want to make the code as readable as possible while keeping the same look.

jquense:
I here that, however as it's a docs page I think we should optimize for maintainability and thats best achieved for me by using react-bootstrap and maybe some css tweaks on the style, all things being equal :)

silvenon:
You're right! I realized how silly I was, I'll switch to React Bootstrap and send you and update soon.

silvenon:
Done! It's not perfect, but it should be much clearer and easier to navigate now.

I strongly suggest forking my CodeSandbox examples into your own account and updating `<iframe>` URLs, to ensure stability.

jquense:
awesome! thanks so much

src/CSSTransition.js
```diff
@@ -109,7 +109,18 @@ const propTypes = {
  * the `example-enter` CSS class and the `example-enter-active` CSS class
  * added in the next tick. This is a convention based on the `classNames` prop.
  *
- * <iframe src="https://codesandbox.io/embed/kw8z6pp9zo?autoresize=1&fontsize=12&hidenavigation=1&moduleview=1" style="width:100%; height:500px; border:0; border-radius: 4px; overflow:hidden;" sandbox="allow-modals allow-forms allow-popups allow-scripts allow-same-origin"></iframe>
+ * ---
+ *
+ * **Note**: `Transition` is platform-agnostic while `CSSTransition`
```
jquense: this is a really great not, i've wanted to add for a while. It might be off tho in the Transition component, since that's the one folks use incorrectly all the time, so may be likely to see it.

I'd also smooth the recommendation out a bit. I think it _is_ ok and often appropriate to use Transition directly for css animations, but it's fraught if you don't know what you're doing.  It might be worth noting that this is a common problem, and if you wnat to use Transtion you need to work around it, or use CSSTransition if it fits your needs.

silvenon: Good point. I modified it a bit and added it to `Transition` page instead.

jquense:
thanks a ton!

Issue #284, opened 2018-02-01
Delay before animation begins


Richardinho:
In the first example given in the documentation there is a noticeable delay between clicking the button and the animation starting. This delay seems to correspond to the duration of the timeout property. Clearly, one would desire the animation to begin straight away, normally, as well as being able to configure a delay in some cases. It's not so obvious in the example because the timeout isn't very large, but when you set it to a bigger number you begin to see the issue.

In my own experiment I set start and end properties for the respective entering and exit states. 
```
const transitionStyles = {
  entering: { left: 0 },
  entered: { left: '100px' },
};
``` 
What then happens is that the start property is immediately set, but that then a delay occurs before the animation kicks in.

I get the desired effect if I set the property to be the same in both states (and set the default to 0), but this doesn't seem right.
```
const defaultStyles = {
  left: 0,
}
const transitionStyles = {
  entering: { left: '100px' },
  entered: { left: '100px' },
};
```
Here is the full code for the my experiment when there is an unwanted delay at the start of the animation:
```
import ReactDOM from 'react-dom';
import React from 'react';
import { Transition } from 'react-transition-group';

const duration = 1600;

const defaultStyle = {
  transition: `left ${duration}ms linear`,
  position: 'absolute',
  top: 200,
  width: '100px',
  height: '100px',
}

const transitionStyles = {
  entering: { left: '0' },
  entered:  { left: '400px'},
};

class Foo extends React.Component {
  constructor () {
    super();
    this.state = {
      in: false,
      duration
    };
    this.onClick = this.onClick.bind(this);
    this.onClickOff = this.onClickOff.bind(this);
  }

  onClick() {
    this.setState({
      in: true,
    });
  }

  onClickOff() {
    this.setState({
      in: false,
    });
  }

  render() {
   return (
     <div>
       <button onClick={this.onClick}>clickOn</button>
       <button onClick={this.onClickOff}>clickOff</button>
       <Transition
        onEntering={() => { console.log('on entering')}}
        onEntered={() => { console.log('on entered')}}
       in={this.state.in}
       timeout={this.state.duration}>
       {(state) => (
         <div style={{
           background: 'blue',
             ...defaultStyle,
             ...transitionStyles[state]
         }}>
         <div>
         <h2>This is Bar</h2>
         </div>
         </div>
       )}
       </Transition>
     </div>
   );
  }
}

ReactDOM.render(
  <Foo/>, document.getElementById('app')
);

```

k-funk:
[...]

PR #313, opened 2018-03-31
Add note about animation timing in the docs

silvenon:
Will hopefully help clarify issues like #284.


silvenon:
I'm not sure I understand, could you elaborate?

jquense:
Sorry, i got lost thinking a bit more wider focus about how we describe Transition. Part of the problem that folks run into like the timeout and reflow issues generally stem from folks thinking the component will implement proper css transitions for them, when really the component is a more of a wireframe that you build implementations on top off. We should probably do a better job of explaining that more broadly.

as a Pr this is GTG and right on the moment 👍

silvenon:
You're right, I'll that as well. 👍

silvenon:
I added the short description of the library where I wrote how react-transition-group does not animate styles. Feel free to merge if you're happy with this.

www/src/pages/index.js
```diff
@@ -24,6 +24,18 @@ class Index extends React.Component {
...
         <h1>React Transition Group</h1>
+        <blockquote>
+          <p>
+            Exposes simple components useful for defining entering and exiting
+            transitions. React Transition Group is not an animation library like{' '}
+            <a href="https://github.com/chenglou/react-motion">React-Motion</a>,
+            it does not animate styles by itself. Instead it exposes transition
+            stages, manages classes and group elements and manipulates the DOM
+            in useful ways, making the implementation of visual transitions up
+            in useful ways, making the implementation of actual visual
```
jquense: duplicate phrase here

silvenon: Fixed.

PR #317, opened 2018-04-03
Lifecycles

jquense:
Update to use new lifecycles


taion:
Did you really have to drop all the semicolons on this PR? 😉

package.json
```diff
@@ -94,9 +94,10 @@
     "eslint-plugin-react": "^6.10.3",
     "jest": "^20.0.4",
     "prettier": "^1.11.1",
-    "react": "^16.0.0",
-    "react-dom": "^16.0.0",
-    "react-test-renderer": "^16.0.0",
+    "react": "^16.3.0",
+    "react-dom": "^16.3.0",
+    "react-lifecycles-compat": "^1.1.3",
```
Andarist: you can upgrade to 2.0 version of this library

src/TransitionGroup.js
```diff
@@ -82,126 +87,75 @@ const defaultProps = {
...
-    this.setState({ children });
+  static getDerivedStateFromProps(
+    nextProps,
+    { children: prevChildMapping, handleExited, firstRender }
+  ) {
+    return {
+      children: firstRender
+        ? getInitialChildMapping(nextProps, handleExited)
+        : getNextChildMapping(nextProps, prevChildMapping, handleExited),
+      firstRender: false,
```
Andarist: `getDerivedStateFromProps` is called also for the first render, so I'm not sure if you have wanted to always return `firstRender: false` here

jquense: The idea is that `firstRender` is intialized to `true`, so if we always set it to false in in gDSFP it will only be true for the first render cycle, does that not track?

Andarist: Oh, I haven't noticed that `firstRender` is not actually used anywhere else than here, but still according to [this](https://twitter.com/dan_abramov/status/981712092611989509) `getDerivedStateFromProps` might be restarted by React, so I believe if that happens by any change (for reasons I dont know 😉) you will end up with already updated value here and unfortunately you will use `getNextChildMapping` instead `getInitialChildMapping`

silvenon:
@jquense I'm sorry about the Prettier configuration, I was hoping that it would work only within the `www` directory because that's where the config is, that's probably what your Prettier editor plugin picked up. Try disabling the plugin only for this project—if you're using VS Code, you can do that by clicking on "Disable (Workspace)":

screen shot 2018-04-13 at 23 11 56

Let me know if I can help more, the point of the change was to painlessly introduce a bit more consistency without getting in the way.

package.json
Hypnosphi: Even to 3.0 now

src/TransitionGroup.js
Andarist: No idea, but I suspect if this hook can be called multiple times before initial render it's very likely that it gets called with updated data from previous calls and not with initial ones - it would be a little bit weird to keep both initial state & calculated state in memory.

Hypnosphi: Ok, then maybe it's worth storing `lastChildren` as state field, call `getInitialChildMapping` in constructor, and `getNextChildMapping` when `nextProps.children !== prevState.lastChildren`

stellarhoof:
I would love this. Errors during testing disrupt the flow and removing strict mode for testing is not an option, since I rely on it to catch early bugs.

vpanjganj:
Ah nice! :) Does this PR resolve Unsafe lifecycle warnings in strict-mode?

Andarist:
@vpanjganj yes, this is the focus of this PR

PR #319, opened 2018-04-14
Move loose-envify and semantic-release to devDependencies

sonaye:


silvenon:
Thanks a lot! Assuming the migration is correct, could you fix the [commitlint error](https://travis-ci.org/reactjs/react-transition-group/builds/366397714?utm_source=github_status&utm_medium=notification)?

silvenon:
@jquense how is it being used? I noticed that the only place it appears is under `browserify.transform` in `package.json`, but I can't even find where `browserify` is being used.

sonaye:
I went through the source before submitting the PR and both packages are not referenced anywhere in the code. loose envify seems related to browserify, as for the error in build, I am not really sure what's causing it.

sonaye:
I will move loose envify back to deps and see if it passes, it shouldn't fix it as it it's not a dep. and the build should pass without it.

jquense:
we should maybe move the linting off the pr branches since we can change the commit when we merge..

thanks for the PR!

sonaye:
@jquense The build failed though even after removing loose envify, something else is causing it. Webpack wont't include it in the bundle since it is not imported anywhere, but eventually it should be removed.

sonaye:
@silvenon Ah! I see, my bad then.

I spent this morning working on [react-animated-transitions](https://github.com/sonaye/react-animated-transitions) an abstraction on top of `react-transition-group` with a pretty straightforward API:

```javascript
import Animated from 'react-animated-transitions';

// animates a child
<Animated>
  <Foo />
</Animated>

// animates a group of children
<Animated items>
  {foos.map(() => <Animated item><Foo /></Animated>)}
</Animated>
```

Neat package you have here, I had to read the code to understand how to build on top of it, thanks for the great work!

jfrej:
So if the build was failing for a different reason, should `loose-envify` be moved to devDependencies now? I can create a PR

silvenon:
@jfrej I'm not sure what we're using it for anyway. I see that it has something to do with browserify, but I don't know what and whether we need to have it at all.

PR #333, opened 2018-05-11
updated dom-helpers from 3.2.0 to 3.3.1 due a bug known crashing in IE11

kafeltz:
update: 
I changed the PR (force-push) from 3.2.1 to 3.3.1 because the dom-helpers author wrongly tagged the version.

### DIFF
--- a/package.json
+++ b/package.json
@@ -43,7 +43,7 @@
   },
   "dependencies": {
     "chain-function": "^1.0.0",
-    "dom-helpers": "^3.2.0",
+    "dom-helpers": "^3.3.1",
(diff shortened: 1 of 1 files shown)


jquense:
Oops sorry, did you mean to submit this to the v1 branch? We aren't really releasing any new updates for v1, PRs should generally be sent against master

kafeltz:
I did in `v1-branch` because it is written in the readme it was still maintaining. I can't migrate to 2.0 right now, it would require me to rewrite severals components because a simple minor issue caused by thirdparty lib.

PR #355, opened 2018-06-16
[v1] update lifecycle methods to be compatible with react@16.3.0

frankwallis:
This removes usage of `componentWillMount` and `componentWillReceiveProps` from the v1-stable branch. I have tried to minimise any other changes. Calculation of `keysToEnter` and `keysToLeave` now happens in `componentDidUpdate` instead of `componentWillReceiveProps` as `getDerivedStateFromProps` does not have access to instance properties.

### DIFF
--- a/src/TransitionGroup.js
+++ b/src/TransitionGroup.js
@@ -25,16 +25,13 @@ class TransitionGroup extends React.Component {
     super(props, context);
 
     this.childRefs = Object.create(null);
+    this.currentlyTransitioningKeys = {};
+    this.keysToEnter = [];
+    this.keysToLeave = [];
 
     this.state = {
       children: getChildMapping(props.children),
     };
-  }
-
-  componentWillMount() {
-    this.currentlyTransitioningKeys = {};
-    this.keysToEnter = [];
-    this.keysToLeave = [];
   }
 
   componentDidMount() {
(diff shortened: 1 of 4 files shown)


taion:
don't we need to polyfill these at least?

frankwallis:
Yes you are right. I wasn't sure if it would be enough to just direct users to the correct version (1.2.1 for react <= 16.2.0 otherwise ^1.3.0). Now I think about it that will cause issues for upstream components which still want to use v1. I will add the polyfills

frankwallis:
I have added the polyfill for `TransitionGroup`, the other components don't need it as they don't use `getDerivedStateFromProps` or `getSnapshotBeforeUpdate`.

src/TransitionGroup.js
```diff
@@ -73,10 +76,6 @@ class TransitionGroup extends React.Component {
       }
     }
 
-    // If we want to someday check for reordering, we could do it here.
```
silvenon: This looks like a TODO comment, we should keep it unless there is a reason to remove it.

silvenon:
Thanks! Let me first try to enable Travis CI for v1. #356

frankwallis:
Any idea when this will be released please?

frankwallis:
Given this is the stable branch perhaps it should go out as a beta version to see if anyone reports issues?

silvenon:
Cc @jquense @taion

frankwallis:
@jquense any chance of a v1 release please?

jquense:
ok i've looked a bit closer at this...the logic in the PR is not the same as the old code. cDU is called more often than cWRP did, and the timing is also very different. I'm not very comfortable releasing this, as we don't really support v1 anymore (though haven't offically deprecated it). Practically maintainers don't use v1 anymore and I don't feel very comfortable in the v1 codebase anymore. I want to push this, but I don't think the benefits necessarily outweigh the risk here. At best this removes some deprecation warnings in strict-mode and we aren't going to upgrade v1 for react v17 so i don't mind so much there. At worst we will break existing folks which is bad :/

taion:
Can we just use the `UNSAFE_foo` methods for now, then?

jquense:
We can, but we need to poly fill it ourselves the compat package doesn't handle that case (intentionally)

frankwallis:
@jquense thanks for looking at this, and I agree it is potentially breaking. If you would like me to raise one with the UNSAFE methods I can do that although there may not be much benefit.

The issue I have is trying to update frankwallis/react-slidedown to V2 and I'm not convinced it is possible due to the lack of callback methods which it uses to cancel previous animations. I will try and find some time to play with it a bit more though.

PR #360, opened 2018-06-27
build: upgrade babel and webpack

jquense:


jquense:
oops should have commented on that. yeah it seemed like a burden for folks contributing, instead lets make the commit the right format during the "squash and merge" step when merging a PR since only we can do it and understand what it should be"

silvenon:
I agree. I added husky in order to make this step easier, but then the problem was that if the commit message was in the wrong format, they had to write it all over again because the commit was aborted! Also, the error message isn't all that helpful, it doesn't show or direct you to any docs about commit types. They were surprisingly hard to find, but I [finally did](https://github.com/angular/angular.js/blob/master/DEVELOPERS.md#type).

Issue #20, opened 2017-02-24
ReactTransitionGroup: Rendering a Single Child


petermikitsh:
**Do you want to request a *feature* or report a *bug*?** Bug

**What is the current behavior?** In the "Rendering a Single Child" [FB](https://facebook.github.io/react/docs/animation.html#rendering-a-single-child) documentation / [readme](https://github.com/reactjs/react-transition-group#rendering-a-single-child), coders are encouraged to pass in prop `component = null` for `ReactTransitionGroup` like so if `childrenArray[0]` is not truthy:

```js
function FirstChild(props) {
  const childrenArray = React.Children.toArray(props.children);
  return childrenArray[0] || null;
}

<ReactTransitionGroup component={FirstChild}>
  {someCondition ? <MyComponent /> : null}
</ReactTransitionGroup>
```

However, if `null` is passed in as `component`, e.g.,

```js
var ReactTransitionGroup = React.addons.TransitionGroup;

var Hello = React.createClass({
  render: function() {
    return <ReactTransitionGroup component={null}/>;
  }
});

ReactDOM.render(
  <Hello />,
  document.getElementById('container')
);
```

The following console error is thrown:

```
Warning: React.createElement: type is invalid -- expected a string (for built-in components) or a class/function (for composite components) but got: null. Check the render method of `ReactTransitionGroup`.
printWarning @ npmcdn.com/react@15.4.2/dist/react-with-addons.js:5276
npmcdn.com/react-dom@15.4.2/dist/react-dom.js:17903 Uncaught Error: Element type is invalid: expected a string (for built-in components) or a class/function (for composite components) but got: null. Check the render method of `ReactTransitionGroup`.
    at invariant (npmcdn.com/react-dom@15.4.2/dist/react-dom.js:17896)
    at ReactCompositeComponentWrapper.instantiateReactComponent [as _instantiateReactComponent] (npmcdn.com/react-dom@15.4.2/dist/react-dom.js:16037)
    at ReactCompositeComponentWrapper.performInitialMount (npmcdn.com/react-dom@15.4.2/dist/react-dom.js:4821)
    at ReactCompositeComponentWrapper.mountComponent (npmcdn.com/react-dom@15.4.2/dist/react-dom.js:4712)
    at Object.mountComponent (npmcdn.com/react-dom@15.4.2/dist/react-dom.js:11542)
[...]

PR #371, opened 2018-07-09
Update TransitionGroup docs

solonaarmstrong:
Fixes https://github.com/reactjs/react-transition-group/issues/20

Clarify what is meant by returning null to TransitionGroup component.

### DIFF
--- a/src/TransitionGroup.js
+++ b/src/TransitionGroup.js
@@ -16,8 +16,8 @@ const propTypes = {
    * `<TransitionGroup>` renders a `<div>` by default. You can change this
    * behavior by providing a `component` prop.
    * If you use React v16+ and would like to avoid a wrapping `<div>` element
-   * you can pass in `component={null}`. This is useful if the wrapping div
-   * borks your css styles.
+   * you can invoke a function and return `null` to `component`. This is useful 
+   * if the wrapping div borks your css styles.
    */
   component: PropTypes.any,
   /**


src/TransitionGroup.js
```diff
@@ -16,8 +16,8 @@ const propTypes = {
    * `<TransitionGroup>` renders a `<div>` by default. You can change this
    * behavior by providing a `component` prop.
    * If you use React v16+ and would like to avoid a wrapping `<div>` element
-   * you can pass in `component={null}`. This is useful if the wrapping div
-   * borks your css styles.
+   * you can invoke a function and return `null` to `component`. This is useful
```
jquense: what function do you mean here?

solonaarmstrong: This issue came up here: https://github.com/reactjs/react-transition-group/issues/20

Passing `component={null}` is not what's intended here. What is intended is:

```
<TransitionGroup component={returnNull} />

function returnNull() {
return null;
}
```

jquense: o I see, the note here is actually talking about a different thing. The comment in #20 refers to the v1 API and is a workaround for rendering only one item or nothing. in the current version (v2) there isn't any need for that workaround anymore because the `Transition` now exists for when you want to render one thing in and out.

 the docs here are correct for what they are saying, which is that TransitionGroup renders a `div` around your children, if you set it to `null` instead it will return just the children with no wrapper, something that is possible in react v16+

solonaarmstrong: Strange. This isn't working for me. In React v16+, it throws an error saying it expected a string, but received null.

PR #383, opened 2018-07-24
fix(CSSTransition): done classname for appearing

### LINKED ISSUE #327: done classname for appearing

bug?

```
    <CSSTransition in appear timeout={timeout} classNames="fade">
        {props.children}
    </CSSTransition>
```

The above case, CSSTransition applies the following classNames in the order.

* `.fade-appear`
* `.fade-appear.fade-appear-active`
* `.fade-enter-done`

Is adding `.fade-enter-done` classname to a component an expected behavior?
I think it would be better the classname will be `.fade-appear-done` instead of `.fade-enter-done`.

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem via https://jsfiddle.net or similar (template: https://jsfiddle.net/m4mb9Lg3/4/).**

https://codesandbox.io/s/7w59nmnn1x

A component is added `.fade-appear-done` instead of `.fade-enter-done`.

koba04:
Fix #327

### DIFF
--- a/src/CSSTransition.js
+++ b/src/CSSTransition.js
@@ -25,6 +25,7 @@ const propTypes = {
    * classNames={{
    *  appear: 'my-appear',
    *  appearActive: 'my-active-appear',
+   *  appearDone: 'my-done-appear',
    *  enter: 'my-enter',
    *  enterActive: 'my-active-enter',
    *  enterDone: 'my-done-enter',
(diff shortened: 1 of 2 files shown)


silvenon:
Thanks for this! There's a discussion in #327 about which classes to apply exactly, feel free to join in.

koba04:
Thanks!

silvenon:
@koba04 thanks for suggesting the change and the PR. 😉

PR #388, opened 2018-08-08
Dont add invalid props to html elements

danielhusar:
When there is html tag as children of `TransitionGroup` before any other component it will throw error:
```
Warning: Unknown props `onExited`, `appear`, `enter`, `exit` on <div> tag. 
Remove these props from the element. For details, see https://fb.me/react-unknown-prop
```
I think its a valid usecase to have html tags before `CSSTransition` or any other component so this fixes it. I dont see also anywhere in documentation that `CSSTransition` must come before any html tag.

### DIFF
--- a/src/utils/ChildMapping.js
+++ b/src/utils/ChildMapping.js
@@ -87,6 +87,8 @@ function getProp(child, prop, props) {
 
 export function getInitialChildMapping(props, onExited) {
   return getChildMapping(props.children, child => {
+    if (typeof child.type === 'string') return cloneElement(child)
+
     return cloneElement(child, {
       onExited: onExited.bind(null, child),
       in: true,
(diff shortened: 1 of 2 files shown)


taion:
Why not just put them outside of the `<TransitionGroup>`, though?

danielhusar:
You can always do that, but I feel like it shouldn't add those props to tags anyway.

silvenon:
What's the use case for using a HTML tag as a `TransitionGroup` child?

danielhusar:
Hmm you can have stuff like 

```js
<TransitionGroup>
 {this.renderHeader()}
 {this.renderBody()}
 {this.renderMeta()}
</TransitionGroup>
```

so you have to remember to change all of them to be actually wrapped in `CSSTransition` or some other component, and not tag. (component could be much more complex)

I think it should really just either be clear that `CSSTransition` or some other component need to be direct child of `TransitionGroup` and throw error or console error in dev if its not, rather than silently adding this props to html tags, or just don't add this props to tags at all.

silvenon:
Yeah, I think we should throw an error or print out a warning if a direct child is a regular HTML element because I don't see a use for it. 👍

Issue #389, opened 2018-08-09
Add transition stage argument to addEndListener


mvasin:
**Do you want to request a *feature* or report a *bug*?**
A feature.

**What is the current behavior?**
I'm trying to use RTG together with GSAP for route transitions. I need GSAP's timeline feature because transitions are pretty involved.

By the end of a GSAP timeline it would be sweet to call a callback to switch to the next transition state.

Currently, `done` callback is only available in `addEndListener` prop. Using it in context of lifecycle hooks (`onEntering` / `onExiting` etc.) requires a wrapper component, that's a complexity on the user side that could be avoided.

As a side note, requiring timeout prop doesn't look right when you come to rely on done callback.

jquense:
I think we've got this covered by #327 ?

> As a side note, requiring timeout prop doesn't look right when you come to rely on done callback.

It's not required! you can leave off the timeout if you use an endListener, it maybe still be beneficial to use both tho depending on what the end listener is, not a problem with GSAP tho

mvasin:
@jquense I don't understand how #327 addresses this case, that issue is about _classes_, and I'm interested in adding an extra `done` callback as parameter to _functions_ `onEnter`, `onEntering`, `onEntered`, `onExit`, `onExiting`. The goal is not to specify duration information twice, in `timeout` and on GSAP timeline.
> It's not required!

Oh, yes indeed, the warning isn't triggered when `addEndListener` is specified.

jquense:
O see, no we won't add that, the callbacks are just that, callbacks. They are informative, not hooks for controlling the the flow. I think what you want is to avoid timeouts entirely and do all the work in the addEndListener, which can setup and control when the end callbacks are fired.

jquense:
I'm happy to add the current t stage to the end listener, I think that's a good addition if you are up for sending a PR. Adding done handlers to the callbacks isn't feasible architecturually tho

mvasin:
OK, I’ll try to come up with a PR.

mvasin:
Closed due to [discovered workaround](https://github.com/reactjs/react-transition-group/pull/396#issuecomment-413791659).

PR #396, opened 2018-08-12
Add transition status param to transition end listener

mvasin:
This is a WIP PR that addresses #389.

I started with tests (in TDD spirit). Any feedback early on is welcome!

### DIFF
--- a/src/Transition.js
+++ b/src/Transition.js
@@ -260,7 +260,7 @@ class Transition extends React.Component {
         this.safeSetState({ status: ENTERED }, () => {
           this.props.onEntered(node, appearing)
         })
-      })
+      }, ENTERING)
     })
   }
 
@@ -284,7 +284,7 @@ class Transition extends React.Component {
         this.safeSetState({ status: EXITED }, () => {
           this.props.onExited(node)
         })
-      })
+      }, EXITING)
     })
   }
 
@@ -322,12 +322,12 @@ class Transition extends React.Component {
     return this.nextCallback
   }
 
-  onTransitionEnd(node, timeout, handler) {
+  onTransitionEnd(node, timeout, handler, status) {
     this.setNextCallback(handler)
 
     if (node) {
       if (this.props.addEndListener) {
-        this.props.addEndListener(node, this.nextCallback)
+        this.props.addEndListener(node, status, this.nextCallback)
       }
       if (timeout != null) {
         setTimeout(this.nextCallback, timeout)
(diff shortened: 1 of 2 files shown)


mvasin:
@jquense I don't mind placing the callback last. But before I'll start to reorder params, please confirm that you still believe that breaking `addEndListener` API is better then adding `done` callback to `onEntering` and `onExiting` hooks (where the callback would naturally go last)?

jquense:
yes we do not want to add any callbacks to onEntering and onExiting, they belong properly in addEndListener

mvasin:
@jquense Seems like I'm mostly done with the PR, please have a look.

Here's the GSAP and updated RTG showcase: https://github.com/mvasin/rtg-gsap-example
I couldn't make it work on codesandbox due to CompuIves/codesandbox-client/issues/1036, so try it locally.

src/Transition.js
```diff
@@ -322,18 +322,14 @@ class Transition extends React.Component {
     return this.nextCallback
   }
 
-  onTransitionEnd(node, timeout, handler) {
+  onTransitionEnd(node, timeout, handler, status) {
     this.setNextCallback(handler)
 
-    if (node) {
```
mvasin: @jquense [GSAP use case](https://github.com/mvasin/rtg-gsap-example/blob/master/src/App.js#L10-L22) does not oblige to provide a DOM node; it can animate a bunch of elements by their classes or ids (and not necessary this `node` element among them), so I removed this check.

mvasin: @jquense According to `git blame`, you are the one who wrote this line. If you have no idea what it stands for, let's remove it.

jquense: I do have idea what it stands for, that's what I'm saying...it prevents edgecase bugs where nodes are detached or mounted/recycled. This isn't an unsure about old code concern, I'm saying that it's a needed check

mvasin: I believe you and am ready to put it back. But it would be a healthy practice to write a test that fails without this conditional and passes with it. Can you write one?

And I'll try to write a test for GSAP use case that fails with this conditional. :) Hopefully we'll find a middle ground.

mvasin: I added tests that ensure addEndListener is triggered even when `<Transition>`'s child renders nothing to the DOM. I think this behaviour is the most versatile; it works well when a child only orchestrates animations and does not render any DOM elements by itself.

It _does_ fail if the `if (node)` conditional is put back.

An interesting issue arose - have a look at [the failing test on Travis](https://travis-ci.org/reactjs/react-transition-group/builds/416226274). If you accept that `<Transition>` can have a child renders nothing to the DOM, I'll investigate it.

jquense: looking good, tho I don't think i understand why the "renders no node" case is important here. It looks like the example you've provided does render nodes, e.g. the page associated with the route.

mvasin: I've used `() => null` component that does not render to DOM and imitates a GSAP component that does timeline animation by ids/classes of other components and does not render to DOM anything on its own.

When `<Transition>` contains `() => null` (named `wrapperDOMlessChild` in the tests), you can see in the failing test that when exiting starts, `<Transition>` is not in `entered` state as expected, but in `exited` instead.

jquense: Can we split that that case into another PR, and keep this focused on adding the status to the endListener? I think we are getting a bit sidetracked on a thing that wasn't the original functional addition

mvasin:
@jquense Alright, I've finished the PR with the idea that a child of `<Transition>` renders something to the DOM. I agree that we can finish off this DOMless child case in a separate PR if needed.

mvasin:
I've discovered a pattern that lets figure out the current transition status using just the current RTG API:
```jsx
<Transition
  in={this.props.in}
  mountOnEnter={true}
  unmountOnExit={true}
  addEndListener={(n, done) => {
    if (this.props.in) {
      console.log("I'm entering!");
      TweenLite.to(...); // trigger done() at the end
    } else {
      console.log("I'm exiting!");
      TweenLite.to(...); // trigger done() at the end
    }
  }}
  >
  <div>Children here</div>
</Transition>
```

It turns out, if you're able to check `<Transition>`'s `in` prop inside `addEndListener`, you're good to go!

Here's the full working example: https://codesandbox.io/s/3yj7qyw9qm

This rises a huge question whether this PR should actually be merged or it's better not to break the API. I lean to the latter unless we'll bump into a use case where figuring out current transition status is a problem.

mvasin:
After using the pattern above a little longer, I've noticed that sometimes what `addEndListener` function receives via `done` argument is not a valid function. Seems like a bug. I'll dig into it.

PR #398, opened 2018-08-18
Throw more descriptive error if no classNames prop

SheaBelsky:
The error currently shown if `classNames` is not specified is not helpful. The way this function is written assumes that if it's not a string, it must be an object, which may not be the case if `this.props.className` is undefined. Now, a `ReferenceError` is thrown if `classNames` is not specified at all.

### DIFF
--- a/src/CSSTransition.js
+++ b/src/CSSTransition.js
@@ -192,25 +192,23 @@ class CSSTransition extends React.Component {
 
   getClassNames = (type) => {
     const { classNames } = this.props;
-    
     let activeClassName;
     let className;
     let doneClassName;
     switch(typeof classNames) {
-      case "string":
+      case 'string':
         className = classNames + '-' + type;
         activeClassName = className + '-active';
         doneClassName = className + '-done';
         break;
-      case "object":
+      case 'object':
         className = classNames[type];
         activeClassName = classNames[type + 'Active'];
         doneClassName = classNames[type + 'Done'];
         break;
       default:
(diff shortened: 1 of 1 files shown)


SheaBelsky:
No, because it isn't [marked as required](https://github.com/reactjs/react-transition-group/blob/master/src/utils/PropTypes.js#L38) there. `oneOfType` only throws an error if a prop passed to the component is not of the shape, but no error is thrown if nothing is passed at all. Adding `.isRequired` to the end of that chain should throw an error if the prop is `undefined` (a prop being `null` is already behind handled as far as I can see.) I'll add that to my branch and test locally to see if it works.

src/CSSTransition.js
```diff
@@ -192,16 +192,23 @@ class CSSTransition extends React.Component {
...
+        activeClassName = className + '-active';
+        doneClassName = className + '-done';
+        break;
+      case 'object':
+        className = classNames[type];
+        activeClassName = classNames[type + 'Active'];
+        doneClassName = classNames[type + 'Done'];
+        break;
+      default:
+        throw new ReferenceError('classNames must be of type object or string, got ' + typeof classNames);
```
jquense: isnt this already handled by the proptype?

SheaBelsky: See further up this PR, it currently isn't. I'll be adjusting this PR to have it be reflected in the PropTypes instead of handling the error here.

Issue #400, opened 2018-08-19
Let the child decide when to switch to the next transition stage


mvasin:
**Do you want to request a *feature* or report a *bug*?**
A feature.

**What is the current and the expected behavior?**
Let's imagine that `<Transition>` renders a child component, and the child component know better when it's done transitioning (i.e. uses `onComplete` GSAP event triggered at the end of animation timeline). It has nothing to do with listening to a CSS `transitionend` event on a particular DOM node, because transition in the child component can involve multiple DOM nodes.

Imagine a dialog:
`<Transition>`: Get ready for your new job, start `entering`!
`<Child>`: OK, I'll unpack my belongings and will tell when I'll be ready to proceed.
`<Child>`: I'm done!
`<Transition>`: Good, you are in `entered` status.
...
`<Transition>`: It's time for you to quit. Take you time as you are `exiting`, I'll wait as long as you need.
`<Child>`: Thanks for waiting, I packed my stuff and ready for departure.
`<Transition>`: Great, now I kick you off. You're `exited`!

At the moment the `react-transition-group` API is not well adapted for such a dialog.

One option we have to set timeout and try to match it with `<Child>`'s transition end time, something that's error-prone and not always possible (think spring animations).

The other option is to assign transition end listener function to `addEndListener` prop. **But the problem here is passing `done` callback from the end listener function to the Child component.** `addEndListener={(_, done) => transitionIsEnded && done()}` also doesn't work because the function is triggered only once, while `transitionIsEnded` is `false`. And we can't save `done` to state because `render()` must be idempotent.

I'm still exploring on what the API should be to allow the dialog above. One idea I have is to provide `<Transition>` with `transitionIsEnded` boolean prop. If it's true, `<Transition>` will switch to the next stage (`entering` -> `entered` or `exiting` -> `exited`). If it's false, it will wait and switch on any of the following conditions:
- `transitionIsEnded` is true
- `addEndListener` triggered `done` callback
- `timeout` expired.

It implies there is an upper component that holds `transitionIsEnded` in state and gives a function to `<Child>` component to toggle that state.

If you have any thoughts on this, please share.

mvasin:
There are happen to be `enter` and `exit` props that I managed use to shortcut directly to the next stage. So looks like the API change is again not needed.

[...]

PR #401, opened 2018-08-19
Add done render prop

mvasin:
WIP to make use case described in #400 possible.

### DIFF
--- a/src/Transition.js
+++ b/src/Transition.js
@@ -139,29 +139,9 @@ class Transition extends React.Component {
     return { transitionGroup: null } // allows for nested Transitions
   }
 
-  static getDerivedStateFromProps({ in: nextIn }, prevState) {
-    if (nextIn && prevState.status === UNMOUNTED) return { status: EXITED }
-    return null
-  }
-
-  // getSnapshotBeforeUpdate(prevProps) {
-  //   let nextStatus = null
-
-  //   if (prevProps !== this.props) {
-  //     const { status } = this.state
-
-  //     if (this.props.in) {
-  //       if (status !== ENTERING && status !== ENTERED) {
-  //         nextStatus = ENTERING
-  //       }
-  //     } else {
-  //       if (status === ENTERING || status === ENTERED) {
-  //         nextStatus = EXITING
-  //       }
-  //     }
-  //   }
-
-  //   return { nextStatus }
+  // static getDerivedStateFromProps({ in: nextIn }, prevState) {
+  //   if (nextIn && prevState.status === UNMOUNTED) return { status: EXITED }
+  //   return null
   // }
 
   componentDidMount() {
--- a/src/Transition.js
+++ b/src/Transition.js
@@ -117,35 +117,68 @@ class Transition extends React.Component {
   constructor(props, context) {
     super(props, context)
 
-    this.nextCallback = null
-
+    let parentGroup = context.transitionGroup
     // In the context of a TransitionGroup all enters are really appears
-    const parentGroup = context.transitionGroup
-    const isInsideTGroup = parentGroup && !parentGroup.isMounting
-    const enterTransitionEnabled = isInsideTGroup ? props.enter : props.appear
+    let appear =
+      parentGroup && !parentGroup.isMounting ? props.enter : props.appear
 
-    // status to transition from on mount (or just keep if statusAfterMount
-    // is the same or null)
-    const ifIn = enterTransitionEnabled ? EXITED : ENTERED
-    const ifOut = props.unmountOnExit || props.mountOnEnter ? UNMOUNTED : EXITED
-    const statusBeforeMount = props.in ? ifIn : ifOut
-    this.state = { status: statusBeforeMount }
+    let initialStatus
 
-    // status to transition to on mount
-    this.statusAfterMount = props.in && enterTransitionEnabled ? ENTERING : null
+    this.appearStatus = null
+
+    if (props.in) {
+      if (appear) {
+        initialStatus = EXITED
+        this.appearStatus = ENTERING
+      } else {
+        initialStatus = ENTERED
+      }
+    } else {
+      if (props.unmountOnExit || props.mountOnEnter) {
+        initialStatus = UNMOUNTED
+      } else {
+        initialStatus = EXITED
+      }
+    }
+
+    this.state = { status: initialStatus }
+
+    this.nextCallback = null
   }
 
   getChildContext() {
     return { transitionGroup: null } // allows for nested Transitions
   }
 
-  // static getDerivedStateFromProps({ in: nextIn }, prevState) {
-  //   if (nextIn && prevState.status === UNMOUNTED) return { status: EXITED }
-  //   return null
+  static getDerivedStateFromProps({ in: nextIn }, prevState) {
+    if (nextIn && prevState.status === UNMOUNTED) {
+      return { status: EXITED }
+    }
+    return null
+  }
+
+  // getSnapshotBeforeUpdate(prevProps) {
+  //   let nextStatus = null
+
+  //   if (prevProps !== this.props) {
+  //     const { status } = this.state
+
+  //     if (this.props.in) {
+  //       if (status !== ENTERING && status !== ENTERED) {
+  //         nextStatus = ENTERING
+  //       }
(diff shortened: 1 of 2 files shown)


mvasin:
@jquense Can you please expand on "In the context of a TransitionGroup all enters are really appears" comment in `src/Transition.js`, and why do you need both `appearStatus` and `initialStatus`?

mvasin:
I figured out `appearStatus` / `initialStatus`, but only after I got rid of hard to follow nested ifs in constructor. Nested ifs are [bad](https://blog.codinghorror.com/flattening-arrow-code/).

jquense:
I'm not sure this API is the right approach. This feels like a change to enable a workaround, instead of adding the right API that you want. It's really not clear that toggling the enabled or disabled phase props should switch the current phase. The reason for these props is to disable or enable which phases the transition group goes through, not a way of controlling the phases.

mvasin:
The more people use any library, the more API grows. I don't want to complicate API with extra options if we can use those we already have.

If you've been in the middle of transition, and found out that you should be going through it in the first place, what's wrong with ending it? It feels like the correct behaviour, shouldn't break other people's code and that's basically all I want.

jquense:
> If you've been in the middle of transition, and found out that you shouldn't be going through it in the first place, what's wrong with ending it?

nothing! but doing that via these props is the part i take issue with, they aren't control flow knobs they are configuration props. They might incidentally short circuit a transition if you toggle them in the middle of one. But that's not likely to cover "allow canceling" well, nor is it a very good api for doing so. I don't actually think tho that is obviously the correct behavior that toggling `enter` midway through an enter should cancel the current transition, i'm more inclined to the think it should do nothing and then affect the next one, since it's much harder for a consumer to have to manage its only transition state and wait for transitions to end if they don't want them canceled

mvasin:
I'm not inclined to introduce the "allow canceling" `enter`/`exit` API, and providing the `done` callback as a second parameter to the `<Transition>` child function should be way easier to implement.

So, I'll do that (I need it for my project anyway), and it's up to you to merge it or not.

mvasin:
@jquense I added the most straightforward callback `() => setState({status: ENTERED})` to render prop (it misses `onEntered` hook for now) and try to make the entering transition termination test pass. The child triggers `done`, but `entered` status never arrives. Your advice would be greatly appreciated.

src/Transition.js
```diff
@@ -361,7 +361,11 @@ class Transition extends React.Component {
     delete childProps.onExited
 
     if (typeof children === 'function') {
-      return children(status, childProps)
+      return children(
+        status,
+        childProps,
+        () => this.setState({status: ENTERED})
```
jquense: this is too simple unfortunately. The done callback really needs to be the same as the actual done callback, otherwise state won't get cleaned up correct. You also need to handle the exiting case as well. I would start by doing something like how setNextCallback works in `onTransitionEnd` and then passing it here from the instance, making sure it gets nulled out when it's not in the entering or exiting state

mvasin:
I've figured out - when the component mounts, `<Transition appear in>` is in `exited` state, and immediate triggering `done` doesn't make sense. If I trigger `done` after making sure current render prop `status` is `entering`, it works. Now I'll build from there.

src/Transition.js
jquense: > relates to race conditions when triggering onEnter*/onExit* hooks

yes and you need to avoid race conditions here as well, you shouldn't be able to call this twice, or after the animation completes

mvasin:
Now `done` works for both entering and exiting, I'm quite happy with that. I'll look into calling hooks next.

Regarding this whole fuss with `safeSetState`: this [stackoverflow thread](https://stackoverflow.com/questions/32903001/) leads to [this article](https://reactjs.org/blog/2015/12/16/ismounted-antipattern.html) that I've read before that says that `this.isMounted()` is an antipattern and only hides problems. Current
```
  componentWillUnmount() {
    this.cancelNextCallback()
  }
```
is exactly that!

I don't want to reproduce this. It's better to find the real problem, fix it, and clean up the code from `safeSetState` crutches. `setState(prevState => {newState}, callback)` callback scheme is safe enough.

jquense:
Hey, I really appreciate you jumping in to contribute and scratch your own itch. However, showing up to a codebase you aren't familiar with and dismissing feedback as not a "best practice" isn't going to win much success. I'm _very_ familiar with the best practices around react, as a maintainer of a lot of core react packages, as well as being maintainer of `react-dom`. If something seems like it might be a bad pattern, feel free to ask, but assume you don't understand all the context instead of assuming the maintainers don't know what they are doing.

> is exactly that!

you have it backwards, the cWUM is cleaning up dangling async callbacks which is exactly what the article suggests to do.

As is it this PR isn't robust enough to merge, the tests you've provided a good start but not sufficient, as they don't cover possible race conditions, don't handle double calls, etc.

test/Transition-test.js
```diff
@@ -131,6 +131,66 @@ describe('Transition', () => {
...
+          {(status, _childProps, doneCB) => (
+            <DoneTrigger status={status} done={doneCB} />
+          )}
+        </Transition>
+      )
+
+      wrapper.setProps({in: false})
+    })
+  })
+
```
jquense: we needs tests, for when components aren't transitioning. Should also make sure that calling done multiple times in a row doesn't break anything.

mvasin:
I didn’t want to offend you and do admit that you understand the context better then I. But looking at the comments in code about safeSetState (like “it should be needed, but somehow it avoids the problems, so let’s use that for now”), I thought it could be one of long standing todos to address. After spending quite a lot of time reading the code, I’ve noticed that the library uses the old context API which is about to be abandoned, has no consistent code style (did you consider prettier?). I thought it’s one of these issues I could try to help with. If you believe the current safeSetState solution is optimal and is required indeed, maybe the comment that it “shouldn’t be needed” should be removed not to confuse people who have less context but still want to contribute.

Regarding the PR, it’s still a WIP, I’ll ping you when it will be ready for review.

Sure I scratch my own itch in the first place, but I’d be happy to improve the library for everybody along the way, and I do appreciate your help with my efforts on both.

test/Transition-test.js
mvasin: I didn’t want to offend you and do admit that you understand the context better then I. But looking at the comments in code about safeSetState (like “it shouldn’t be needed, but somehow it avoids the problems, so let’s use that for now”), I thought it could be one of long standing todos to address. After spending quite a lot of time reading the code, I’ve noticed that the library uses the old context API which is about to be abandoned, has no consistent code style (did you consider prettier?). I thought it’s one of these issues I could try to help with. If you believe the current safeSetState solution is optimal and is required indeed, maybe the comment that it “shouldn’t be needed” should be removed not to confuse people who have less context but still want to contribute.

Regarding the PR, it’s still a WIP, I’ll ping you when it will be ready for review.

Sure I scratch my own itch in the first place, but I’d be happy to improve the library for everybody along the way, and I do appreciate your help with my efforts on both.

mvasin:
Thanks for the code comments, it will help improving the code. If you’re sure the dangling callbacks cancellation workaround is required, I’ll use it.

src/Transition.js
```diff
@@ -337,6 +337,14 @@ class Transition extends React.Component {
     }
   }
 
+  doneRenderPropCallback = () => {
+    this.setState(({status: prevStatus}) => {
```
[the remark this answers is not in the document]
mvasin: As I try to exploit my PR early on in my project and bump into the exception I added, I start to grasp the race condition problem we have.

If we pass the child direct control over `<Transition>` state with `setState`, it thinks it can use that handle any time. Then occasionally `in` prop on `<Transition>` is toggled by the upper component, and that turns internal `status` of `<Transition>` in a different state. If after that the child component shows up with its handle and tries to 'end transition', that becomes irrelevant as Transition's internal `status` is not in the same state the child expects.

Indeed, it has nothing to do with best/worst practices, it's inherent to the situation where (in this case) we try to control `<Transition>` state machine from two separate ends (`in` prop and `done` render prop).

I'll try to model these situations in tests.

> ensure that dangling listeners can be cleaned up correct on unmount

By unmount do you mean the `Transition` component unmount or `unmounted` status in `Transition` state?

PR #409, opened 2018-09-14
Uprgade to stable babel 7

TrySound:
Tweaked a bit config. Removed custom config in favour of 3 plugins which
shows more clear how code looks like.


TrySound:
Do you mean browserlist config? Current setting transpiles everything to be uglifiable.

jquense:
Sure, that's different from what they were. We want to maintain the same browser support. I would keep this simple and upgrade the preset instead

TrySound:
Preset will prevent me from further improvements.

Don't you think it's too opinionated? When did we start to distribute untranspiled code for frontend? In any case, I see everything is transpiled in the latest release.

https://unpkg.com/react-transition-group@2.4.0/utils/ChildMapping.js

jquense:
I mean it's opinionated in the sense that I'm the maintainer and it's the settings that support my use cases/browsers and reduce my maintainabce burden across projects

> When did we start to distribute untranspiled code for frontend

We don't, I'm not sure what you mean...my only concern here is you replaced the existing config with different one that isn't equivalent. Specifically the default settings of preset env will produce tranpiled code that has a different support matrix than the existing one.

TrySound:
Okay. Looks good?

TrySound:
@jquense Is this okay for you? I would like to continue improvements.

jquense:
Hey @TrySound i've rethought this. I spend too much time dealing with babel quirks and bugs to maintain a separate config for every project i own. It's a significant benefit that i can fix bugs when they arise once and then upgrade the preset in projects. I realize this can make the those projects harder to contribute to and that is not ideal, however i need to also protect my limited time so it's spent doing things besides dealing with build tooling on projects. 

If you still want to make improvements please do so as additional configuration in the babelrc or if it makes sense to upstream it in the preset. I would propose what you intend to do tho first before implementing anything to make sure it'll get in. Sorry for the inconvenience and I really do appreciate that you take time to try and improve the build output of this and other projects i work on thanks!

PR #416, opened 2018-09-27
Remove loose-envify from dependency + cleanup unwanted devDependencies

sibiraj-s:
* removed `loose-envify` from dependency. I can't see how `loose-envify` is a dependant, I have removed it. If needed for browserify it should be in devDependency. Seems browserify is not used too. If that is some how still needed I will re add it.
* removed `husky` and `rimraf`

### DIFF
--- a/package.json
+++ b/package.json
@@ -55,7 +55,6 @@
   },
   "dependencies": {
     "dom-helpers": "^3.3.1",
-    "loose-envify": "^1.4.0",
     "prop-types": "^15.6.2",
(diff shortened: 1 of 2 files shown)


sibiraj-s:
Thanks @jquense for the info.

Also >hucky is also used, as it handles precommit hooks

It seems you removed it #360  _. I have removed that after I saw that. there is no use of any hooks now.

I have updated the PR.

sibiraj-s:
Okay. My bad closed PR's won't update. 👍

PR #437, opened 2018-12-15
Use React.createRef() to track child nodes (Remove ReactDOM dependency)

kjkta:
Here's a potential solution to removing ReactDOM peer dependency.

### DIFF
--- a/package.json
+++ b/package.json
@@ -50,8 +50,7 @@
     ]
   },
   "peerDependencies": {
-    "react": ">=15.0.0",
-    "react-dom": ">=15.0.0"
+    "react": ">=16.3.0"
   },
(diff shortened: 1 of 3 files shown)


taion:
there’s no guarantee the child will be a DOM node. among other things

kjkta:
How does using refs to identify a node (if there is one) differ to using ReactDOM.findNode?

Miklet:
`ref` can be a node but also can be a component's instance. Using [`ReactDOM.findNode`](https://reactjs.org/docs/react-dom.html#finddomnode) always return underlaying node (unless it renders null or false).  Passing explicitly a ref would guarantee it but it would break API.

PR #448, opened 2019-01-13
Strip custom prop-types in production

TrySound:
Minified size with bundled dependencies before
17.53 KB
After
16.66 KB

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,6 +21,13 @@
     "travis-deploy-once": "travis-deploy-once",
     "semantic-release": "semantic-release"
   },
+  "size-limit": [
+    {
+      "gzip": false,
+      "path": "lib/index.js",
+      "limit": "16.66 KB"
+    }
+  ],
   "repository": {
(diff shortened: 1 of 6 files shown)


TrySound:
Yep

TrySound:
@jquense Is this good now?

oliviertassinari:
-200 B gzipped https://bundlephobia.com/result?p=react-transition-group@2.5.3 🚀

jquense:
hmm we are still including the prop-types library tho. seems like the import isn't getting stripped

oliviertassinari:
@jquense It's not. I wish the babel plugin was doing such.

oliviertassinari:
It's another gain potential. https://bundlephobia.com/result?p=prop-types@15.6.2. However, I'm not sure how we can make it work with ESM modules. Conditionally importing modules? I doubt we can 🤔.

TrySound:
It's because of old context api I guess

Issue #429, opened 2018-10-25
StrictMode warning for findDOMNode and legacy context API in Transition as of React 16.6


TAGraves:
**Do you want to request a *feature* or report a *bug*?**
Report a bug

**What is the current behavior?**
When using `<Transition />` inside of `<React.StrictMode />`, warnings are logged about `findDOMNode` and the legacy context API (as of React 16.6)

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem via [CodeSandbox](https://codesandbox.io/).**
View the console at https://codesandbox.io/s/y0r3kl6x1j. On mount, you'll see:
```
Warning: Legacy context API has been detected within a strict-mode tree
```

and after clicking "Toggle in Prop", you'll see:
```
findDOMNode is deprecated in StrictMode. findDOMNode was passed an instance of Transition which is inside StrictMode. Instead, add a ref directly to the element you want to reference.
```

**What is the expected behavior?**
No warning should be logged.

**Which versions, and which browser / OS are affected by this issue? Did this work in previous versions?**
These warnings are new as of React 16.6.

jquense:
Thanks! I wouldn't consider this a bug since it's a deprecation warning but we'll do what we can to use refs if feasible

edmorley:
Please can this be reopened until it's fixed? :-)

The warning makes using `<StrictMode>` very noisy, which effectively blocks us using it to prevent other issues.

jquense:
The warning can't stop transition s it's just a console.log not a thrown error

ekohanyi:
It might be something else but we noticed the transitions weren't firing and then when we turned off strict mode it started animating properly again, so seems like something is getting stopped up related to it

Kovensky:
StrictMode doesn't _only_ print warnings, it also double-renders all components in development.

ekohanyi:
@Kovensky Ohh okay, good to know, that must be what's causing the issue then. Thank you! I learned something new today haha

Iverson:
Is there any progress here? It's really painful to work with console in `<StrictMode>`.

sseppola:
I haven't been able to wrap my head around all the considerations this lib takes, but to move the conversation forward:

### 1) Are there any specific challenges to use the new context api?
From what I could gather by browsing the source code, maybe it could cause problems for nested transitions? (https://github.com/reactjs/react-transition-group/blob/master/src/Transition.js#L150)

[...]

PR #457, opened 2019-02-14
feat(Transition): Add findDOMNode prop

eps1lon:
Useful for strict mode compatibility. Part of #429.

Started out with something along the lines of a `skipDOMNode` but this would've been equivalent to `<Transition findDOMNode={() => undefined} />`. Current solution allows more flexibility by preserving the call signature of the state transition callbacks while being strict mode compatible with regards to `findDOMNode` deprecation.

Not sure how/if I should update the docs.

### DIFF
--- a/src/ReplaceTransition.js
+++ b/src/ReplaceTransition.js
@@ -28,7 +28,7 @@ class ReplaceTransition extends React.Component {
     const child = React.Children.toArray(children)[idx];
 
     if (child.props[handler]) child.props[handler](...originalArgs)
-    if (this.props[handler]) this.props[handler](findDOMNode(this))
+    if (this.props[handler]) this.props[handler](this.props.findDOMNode(this))
   }
 
   render() {
@@ -39,6 +39,7 @@ class ReplaceTransition extends React.Component {
     } = this.props;
     const [first, second] = React.Children.toArray(children);
 
+    delete props.findDOMNode;
     delete props.onEnter;
     delete props.onEntering;
     delete props.onEntered;
@@ -76,6 +77,11 @@ ReplaceTransition.propTypes = {
 
     return null;
   },
+  findDOMNode: PropTypes.func,
 };
+
+ReplaceTransition.defaultProps = {
+  findDOMNode,
+}
 
 export default ReplaceTransition;
--- a/src/Transition.js
+++ b/src/Transition.js
@@ -209,7 +209,7 @@ class Transition extends React.Component {
     if (nextStatus !== null) {
       // nextStatus will always be ENTERING or EXITING.
       this.cancelNextCallback()
-      const node = ReactDOM.findDOMNode(this)
+      const node = this.props.findDOMNode(this);
 
       if (nextStatus === ENTERING) {
         this.performEnter(node, mounting)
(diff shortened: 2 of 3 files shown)


taion:
Is this the path forward that we want? I'd imagine that https://github.com/reactjs/rfcs/pull/97 would supersede this in a nicer way.

eps1lon:
> Is this the path forward that we want? I'd imagine that [reactjs/rfcs#97](https://github.com/reactjs/rfcs/pull/97) would supersede this in a nicer way.

At which point we drop `ReactDOM.findDOMNode` and use that rfc. This approach doesn't contradict this. It is a bridge until a better solution is found. That RFC however isn't even accepted yet or any implementation planned so I don't get my hopes up.

taion:
It'd be a breaking change to remove this API, though.

jquense:
I think in the short term we probably just want to switch to full ref props, and cut a breaking release. I'm not in a huge rush for that tho, but we can make breaking releases. The reticence here to add new public api's is that this is all to work around a warning in strict mode, not an actual bug. It's just generally lower priority to fix, especially when weighed against public API changes required

eps1lon:
> It'd be a breaking change to remove this API, though.

Sure. But I wasn't the one to suggest a (maybe) future feature of react.

> The reticence here to add new public api's is that this is all to work around a warning in strict mode, not an actual bug.

I see this argument all the time: "It's just a warning". This is just not realistic for library authors. Especially if this is a transitive dependency. There is currently no way in react development to filter out deprecation warnings, prop types warnings or internal react warnings. These are all the same. A runtime warning in transitive dependencies is considered a bug by man people. It adds at least a lot of noise that is indistinguishable from actual errors in the console.

> [...] in strict mode, not an actual bug.

Strict mode **and** concurrent mode. I'm not sure how findDOMNode behaves in that mode. Might be a bug there. It's especially a bug while testing. `mocha` will not let tests pass for any warning that is issued by react because they use console.error.

> API changes required

It's an addition with virtually zero runtime overhead and (fairly certain) no bundle size increase. If you're not interested in being strict or concurrent mode ready then I can close this.

jquense:
> These are all the same. A runtime warning in transitive dependencies is considered a bug by man people. It adds at least a lot of noise that is indistinguishable from actual errors in the console.

Believe me, I HEAR you on this, I maintain a ton of react libraries. The plain fact tho is this is not a bug and it doesn't benefit us as maintainers to accept the incorrect claim that they are from some users. React, itself holds this position, is explicitly about how these are not bugs. It's not a tenable position for maintainers to have to drop everything on every minor and patch version of react to silence new warnings, 

> There is currently no way in react development to filter out deprecation warnings, prop types warnings or internal react warnings.

These are strict mode warnings so are entirely opt in. 

> If you're not interested in being strict or concurrent mode ready then I can close this.

We very interested in being ready! I even offered an approach above. This is not about refusing to the fix the problem, but in fixing it in a sustainable way and with a look to future proofing, API churn reduction etc. We **want** to fix this, but there is currently no pressing rush to do so that we need to pick a less optimal solution.

eps1lon:
Sounds fair to me. Let's say we might be able to determine in a component if we're in a strict mode tree. Would you be open to skip the `findDOMNode` call if a given `Transition` component is a child of `React.StrictMode`?

PR #462, opened 2019-02-26
Add appear timeout

dimensi:
Hi, I had a case when I needed the appear timeout, I saw in the code that the opportunity was almost implemented, but unfinished. I hope that I completed this feature.

### DIFF
--- a/src/Transition.js
+++ b/src/Transition.js
@@ -437,8 +437,11 @@ Transition.propTypes = {
    * timeout={{
    *  enter: 300,
    *  exit: 500,
+   *  appear: 500,
    * }}
    * ```
+   *
+   * If the value of appear is not set, then the value from enter is taken.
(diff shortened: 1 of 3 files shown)


src/Transition.js
```diff
@@ -212,7 +212,7 @@ class Transition extends React.Component {
     if (timeout != null && typeof timeout !== 'number') {
       exit = timeout.exit
       enter = timeout.enter
-      appear = timeout.appear
+      appear = timeout.appear !== undefined ? timeout.appear : enter
```
jquense: let's add a TODO here to remove this fallback on the next major bump

dimensi: added

jquense:
thanks!

dimensi:
thanks for fast response.

Issue #460, opened 2019-02-19
use timeout props without exit time


LiuuY:
**Do you want to request a *feature* or report a *bug*?**

bug, maybe?

**What is the current behavior?**

if I use timeout props without exit time, eg, 
```javascript
<CSSTransition
  key={id}
  timeout={{
    enter: 300,
    // no exit time
  }}
  classNames="fade"
>
```

https://codesandbox.io/s/wnjp84n125

Click the red X button, after the animation, the element that should (?) be removed from DOM tree is still there.

Is this a feature or a bug?

![image](https://user-images.githubusercontent.com/14286374/53026910-d71d2780-349e-11e9-9a43-abb8f3b7e389.png)

version: 2.2.1

silvenon:
I think this is unexpected behavior. I think the expected behavior should be to raise an error or warning, because I don't know if there is a sensible default value we could add for `exit`. (Maybe `0`?)

silvenon:
Solved with #464. `exit` timeout now defaults to `0`, which I consider intuitive behavior.

PR #464, opened 2019-02-28
fix bug from issue #460

dimensi:
I found an https://github.com/reactjs/react-transition-group/issues/460 with a bug, since the bug is simple, I decided to correct the error.
The error was that we did not take into account the case when there is a node and there is no timeout.
I also set the enter and exit values in prop-types as reqiured.

### DIFF
--- a/src/Transition.js
+++ b/src/Transition.js
@@ -326,7 +326,7 @@ class Transition extends React.Component {
     this.setNextCallback(handler)
 
     const doesNotHaveTimeoutOrListener = timeout == null && !this.props.addEndListener
-    if (!node || isNotHaveTimeoutOrListener) {
+    if (!node || doesNotHaveTimeoutOrListener) {
       setTimeout(this.nextCallback, 0)
       return
     }
@@ -449,7 +449,7 @@ Transition.propTypes = {
    *
    * If the `enter` or `exit` value is `null` or `undefined`, then the timer is set to `0`
    *
-   * @type {number | { enter: number, exit: number, appear?: number }}
+   * @type {number | { enter?: number, exit?: number, appear?: number }}
    */
   timeout: (props, ...args) => {
     let pt = timeoutsShape
--- a/src/Transition.js
+++ b/src/Transition.js
@@ -325,15 +325,18 @@ class Transition extends React.Component {
   onTransitionEnd(node, timeout, handler) {
     this.setNextCallback(handler)
 
-    if (node) {
-      if (this.props.addEndListener) {
-        this.props.addEndListener(node, this.nextCallback)
-      }
-      if (timeout != null) {
-        setTimeout(this.nextCallback, timeout)
-      }
-    } else {
+    const isNotHaveTimeoutOrListener = timeout == null && !this.props.addEndListener
+    if (!node || isNotHaveTimeoutOrListener) {
       setTimeout(this.nextCallback, 0)
+      return
+    }
+
+    if (this.props.addEndListener) {
+      this.props.addEndListener(node, this.nextCallback)
+    }
+
+    if (timeout != null) {
+      setTimeout(this.nextCallback, timeout)
     }
   }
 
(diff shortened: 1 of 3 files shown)


dimensi:
@jquense ?

silvenon:
Reviewing your PR now. 👀

silvenon:
Concerning prop types, you can take this opportunity to actually fix them:

```js
PropTypes.oneOfType([
  PropTypes.number,
  PropTypes.shape({
    enter: PropTypes.number,
    exit: PropTypes.number,
    appear: PropTypes.number,
  })
]).isRequired
```

I moved `isRequired` to the correct place, now people who forget to add the `timeout` prop will finally get a warning.

I'm honestly not sure what `isRequired` does within `oneOfType` 🤔

src/Transition.js
```diff
@@ -442,9 +445,11 @@ Transition.propTypes = {
    * }}
    * ```
    *
-   * If the value of appear is not set, then the value from enter is taken.
+   * If the value of `appear` is not set, then the value from enter is taken.
+   *
+   * If the `enter` or `exit` value is `null` or `undefined`, then the timer is set to `0`
    *
-   * @type {number | { enter?: number, exit?: number }}
+   * @type {number | { enter: number, exit: number, appear?: number }}
```
silvenon: Make `enter` and `exit` optional again. And thanks for adding `appear` timeout to the docs too! 👏

src/utils/PropTypes.js
```diff
@@ -5,8 +5,8 @@ export const timeoutsShape =
     ? PropTypes.oneOfType([
         PropTypes.number,
         PropTypes.shape({
-          enter: PropTypes.number,
-          exit: PropTypes.number,
+          enter: PropTypes.number.isRequired,
+          exit: PropTypes.number.isRequired,
           appear: PropTypes.number,
         }).isRequired
       ])
```
silvenon: Move `isRequired` from `shape` to `oneOfType`. Totally optional, but would be awesome 😎

test/Transition-test.js
```diff
@@ -438,5 +438,48 @@ describe('Transition', () => {
...
+            })
+          }}
+        />
+      ).instance()
+
+      expect(wrapper.getStatus()).toEqual(ENTERED)
+      expect(ReactDOM.findDOMNode(wrapper)).toExist()
+
+      wrapper.setState({ in: false })
+    })
```
silvenon: Simplify tests to measure that in the absence of `enter` and `exit` the component immediately entered and exited respectively.

src/Transition.js
```diff
@@ -325,15 +325,18 @@ class Transition extends React.Component {
...
 
-    if (node) {
-      if (this.props.addEndListener) {
-        this.props.addEndListener(node, this.nextCallback)
-      }
-      if (timeout != null) {
-        setTimeout(this.nextCallback, timeout)
-      }
-    } else {
+    const isNotHaveTimeoutOrListener = timeout == null && !this.props.addEndListener
```
silvenon:
```suggestion
    const doesNotHaveTimeoutOrListener = timeout == null && !this.props.addEndListener
```

src/utils/PropTypes.js
dimensi: Ok, you're right, i remove isRequired.

src/Transition.js
dimensi: Ok

dimensi:
> Could you give more details about what your tests are proving? By "at once" do you mean "immediately"? How is time measured here?

yes, i don't know english, i use google translate for find a word.

test/Transition-test.js
dimensi: @silvenon i update test, but i not sure i write it right.

silvenon:
Awesome, thank you! 🏆

dimensi:
@silvenon thx for your review. Are you sure my tests right writed?

silvenon:
@dimensi yep, they look good to me. I didn't even think to combine both functionalities in the same test. We might later split that into two tests because they are two different features.

PR #465, opened 2019-03-05
chore: fix Gatsby build

silvenon:
This is a mix of updates, sorry about that, I couldn't really narrow down why Gatsby was failing for me (I don't know if anyone else was experiencing this), it seems to have been a combination of issues which were taking too long to isolate. Anyway, now our Gatsby dependencies are out of beta and we're using official Gatsby Babel preset, so things should be more stable in the future.

Also, I migrated to latest react-bootstrap, very easy migration so 👍 to the team 😉

Can't wait to start upgrading docs with better examples after this…

### DIFF
--- a/www/.babelrc.js
+++ b/www/.babelrc.js
@@ -1,1 +1,5 @@
-module.exports = require('../.babelrc');
+module.exports = {
+  presets: [
+    'babel-preset-gatsby',
+  ],
+};
--- a/www/package.json
+++ b/www/package.json
@@ -1,4 +1,5 @@
 {
+  "private": true,
   "name": "react-transition-group-docs",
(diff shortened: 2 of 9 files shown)


silvenon:
Since this is just a change to building the docs, I'll go ahead and merge this.

dimensi:
@silvenon what's happen with documentations on https://reactcommunity.org/react-transition-group/ ? It is not updated after releases, is it because skip ci?

silvenon:
@dimensi good question. I think it needs to be deployed manually with the current setup, I'm looking into it and will deploy it one way or another 😉 

But I still want to brush up the docs content before deploying. Feel free to suggest any additions. 😉

PR #468, opened 2019-03-12
feat(Transition): Only call findDOMNode if necessary

eps1lon:
Another approach to #457. It only calls findDOMNode if any of the callbacks has the node argument specified. 

This is only exploratory since seemingly unrelated implementation had to be change. Details at the affected line.


src/Transition.js
```diff
@@ -326,7 +326,7 @@ class Transition extends React.Component {
     this.setNextCallback(handler)
 
     const doesNotHaveTimeoutOrListener = timeout == null && !this.props.addEndListener
-    if (!node || doesNotHaveTimeoutOrListener) {
+    if (doesNotHaveTimeoutOrListener) {
```
eps1lon: Has its own commit (87832e91489d046012006c16cad58bd216cb40b2). No test broke upon removing this so I need some guidance to understand this condition.

jquense:
this is cool, the problem nowdays with this approach tho is destructuring messings with the observable arity of the function e.g.

```js
function foo(...args) {
  console.log(args
}
foo.length // 0
```

eps1lon:
> this is cool, the problem nowdays with this approach tho is destructuring messings with the observable arity of the function e.g.
> 
> ```js
> function foo(...args) {
>   console.log(args
> }
> foo.length // 0
> ```

It was to good to be true. This would be a breaking change then at which point we might as well remove the call completely.

Any thoughts about https://github.com/reactjs/react-transition-group/pull/468#discussion_r264817184?

src/Transition.js
dimensi: https://github.com/reactjs/react-transition-group/pull/464

eps1lon: That PR did not introduce the evaluation of `node`.

eps1lon:
Fixed with https://github.com/reactjs/react-transition-group/pull/559

PR #469, opened 2019-03-12
fix(Transition): Stale nodes in transition callbacks

eps1lon:
- 3f3c1e36292e97803250c05c842dfcaea86745e5 for the current behavior
- 7003c58c0fc52d2b31b985dae7ce0c532fd6ebfe for the fix

The issues is that `findDOMNode` is only called a single time after `componentDidUpdate` and the return value is used for every transition callback that is triggered by this update. If the child however changes its DOM node while transitioning this value is stale.


iamandrewluca:
I think this can be closed. See https://github.com/reactjs/react-transition-group/pull/559

eps1lon:
> @silvenon I think this can be closed. See #559

I'll rebase and see if the bug persists. But I suspect this still happens for Transition using findDOMNode.

iamandrewluca:
As a note. When user is passing `nodeRef` props, `node` should not be passed to callback functions.

iamandrewluca:
> The issues is that findDOMNode is only called a single time after componentDidUpdate and the return value is used for every transition callback that is triggered by this update. If the child however changes its DOM node while transitioning this value is stale.

I'm looking into code and it seems after #559 statement above is false one?!
Or I don't understand exactly what you mean.

eps1lon:
> this.getNode does not add too much value thinking

I could inline it sure. But it's used often enough that I think abstracting it is warranted.

Issue #339, opened 2018-05-18
[Feature]: Transition Modes for First-Class Sequential Enter/Leave (instead of Parallel)


tsiq-swyx:
**Do you want to request a *feature* or report a *bug*?**

Feature 

**What is the current behavior?**

Currently, entering and leaving elements run in parallel. You can choose to use CSS to overlay the entering element on top of the leaving element, and you can manually keyframe transitions such that one starts after the other, but this is gnarly as heck (and quite frankly I couldn't find a blogpost on it). In Vue [Transition Modes](https://vuejs.org/v2/guide/transitions.html#Transition-Modes) you can just specify `mode="out-in"` and you're done. I think this is a common enough pattern (and best handled by the library instead of in userland/in css).

**What is the expected behavior?**

See Vue Transition modes https://vuejs.org/v2/guide/transitions.html#Transition-Modes
and see also [Sarah Drasner's page transitions app in Vue](https://page-transitions.com/) vs [its React clone](https://page-transitions-react-travelapp.netlify.com/)

Potentially related issue: https://github.com/reactjs/react-transition-group/issues/73

jquense:
i do want to add this, and I think it's totally feasible someone just has to do it. I think i have a started POC adding them to the ReplaceTransition which is a little more akin to the VUE transition. I'll see if i can dig it up

tsiq-swyx:
cool. idk if i can commit to doing a PR as im pretty full up, its just a thing that ive wished for that came up when Sophie Alpert asked for easy wins

jquense:
here's the branch if anyone is interested https://github.com/reactjs/react-transition-group/tree/modes-spike

tsiq-swyx:
lol Kyle Mathews just pointed me to https://github.com/cheapsteak/react-transition-group-plus and it has exactly what we need.. only problem is they forked off 2 years ago...

igormigas:
So... any chance to get the "out-in" feature running in this package?

IGassmann:
@tsiq-swyx What's the best way so far you've figured to achieve this?

PR #470, opened 2019-03-12
feat: add SwitchTransition component

dimensi:
Hi @jquense and @silvenon, I came from vue and when I didn’t find the transition modes for the transition component, I started biting my elbows and worked with animations using `position: absolute`. I put together a fist and wrote a component that adds transition modes.
I saw [this](https://github.com/reactjs/react-transition-group/issues/339) and that the @jquense wants to [add it](https://github.com/reactjs/react-transition-group/issues/339#issuecomment-390315102). But the idea of the @jquense is that you need to pump through exactly the same component, I was not understood.
Here is a live example of this component https://codesandbox.io/s/50p5wr3jwl.
My idea is that I essentially repeat the behavior of the vue transition. If you change the key, the animation will play, if you change the components, but do not touch the keys, the component will be updated, but the animation will not play.
I have not added tests or jsdoc yet, because I want to get feedback first.


dimensi:
@jquense, @silvenon  ?

jquense:
Hi @dimensi this looks great! sorry for the delay folks are working on stuff in their spare time and we're all a bit busy right now, so sorry for the long gap. Hoping i can give this a proper review soon. thanks a ton for doing this!

dimensi:
Sorry, @jquense, i really want release this feature.

jquense:
Looks good @dimensi can you add some tests and documentation?

dimensi:
@jquense, I've added documentation and tests

dimensi:
@jquense, i update switch for context api, but now can't create sandbox with new version, because transitiongroup context now not exported.

jquense:
@dimensi you should be able to cherry pick it `import TransitionGroupContext from 'react-transition-group/TransitionGroupContext'`

dimensi:
image
codesandbox not see /esm or cjs forders

jquense:
thanks! sorry for the long delay :\

dimensi:
I leave here working example: https://codesandbox.io/s/50p5wr3jwl

PR #471, opened 2019-03-12
breaking: Use stable context API

eps1lon:
`React@16.6.0` allows usage of static `contextType`. With `16.0.0` we would need a higher-order component that puts the context value into props since `Transition` needs to have access to the context in its constructor.

Partial fix for #429  

### breaking change
```diff
// package.json
-"react": "^15.0.0",
+"react": "^16.6.0",
-"react-dom": "^15.0.0", 
+"react-dom": "^16.6.0", 
```


TrySound:
Duplicate of https://github.com/reactjs/react-transition-group/pull/427

eps1lon:
> Duplicate of #427

Same goal, different implementation. This one uses `contextType` instead of a hoc and doesn't have 3rd party dependencies.

But yes it would've been nice to link this from the the corresponding issue.

package.json
```diff
@@ -50,8 +50,8 @@
     ]
   },
   "peerDependencies": {
-    "react": ">=15.0.0",
-    "react-dom": ">=15.0.0"
+    "react": ">=16.6.0",
```
oliviertassinari: We can kill `react-lifecycles-compat`, now please!
https://bundlephobia.com/result?p=react-lifecycles-compat@3.0.4

Vodzo:
Could this be released as @3 or even beta? Current legacy context implementation has issues when used with latest react-router and react-redux both of which are using the new Context API. This pull request fixes those issues  I'm seeing in my code

jquense:
we've got semantic release now, let's (talking mostly to myself) not be afraid of bumping integers

TrySound:
@jquense So fast. I expected esm support to be released in v3 as major change.

PR #472, opened 2019-03-13
fix(package.json): mark react-transition-group as side-effect free for webpack tree shaking

Setsun:
## What
Add `"sideEffects": false` to `package.json` to hint to `webpack` based workflows that the package is side effect free. This allows `react-transition-group` to allow tree-shaking of its modules.

Reference: https://webpack.js.org/guides/tree-shaking/#mark-the-file-as-side-effect-free

### DIFF
--- a/package.json
+++ b/package.json
@@ -123,5 +123,6 @@
     "transform": [
       "loose-envify"
     ]
-  }
+  },
+  "sideEffects": false
 }


billyjanitsch:
@Setsun @silvenon Unfortunately this doesn't have any effect because react-transition-group only ships a CommonJS build. It would have to also ship an ESM build (and point to it with the `module` field in package.json) for `{sideEffects: true}` to do anything. Module-level tree-shaking isn't possible with CommonJS.

Setsun:
Ah interesting, that's good to know. This may warrant a follow up PR

silvenon:
Actually, #455 already adds this, we just haven't merged it yet. I was really busy and honestly not really knowledgeable about modules, but now I'll take some time to properly review and merge it if it's all good. 🙂

PR #481, opened 2019-04-01
Add support for empty classNames

danielhusar:
Hey!

I think it would be quite usefull if `classNames` prop would be optional, and it would still work without it.

In our use-case we are using it together with styled components and it's quite annoying to add those prefixes all the time, when component styles are already scoped.

Also if you pass empty string, it adds classes beginning with `-`.

### DIFF
--- a/src/CSSTransition.js
+++ b/src/CSSTransition.js
@@ -71,6 +71,9 @@ const removeClass = (node, classes) => node && classes && classes.split(' ').for
  * `*-active` classes represent which styles you want to animate **to**.
  */
 class CSSTransition extends React.Component {
+  static defaultProps = {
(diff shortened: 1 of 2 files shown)


src/CSSTransition.js
```diff
@@ -140,9 +143,10 @@ class CSSTransition extends React.Component {
 
   getClassNames = (type) => {
     const { classNames } = this.props;
+    const prefix = classNames ? classNames + '-' : '';
```
jquense: i'd also check that it's a string to avoid concating an object to a string unnecessarily

danielhusar: done!

jquense:
thanks!

PR #486, opened 2019-04-11
Updated to use refs rather than `findNode`

thaggie:
I have a situation where `findNode` was throwing an error because the node wasn't mounted yet.

### DIFF
--- a/src/Transition.js
+++ b/src/Transition.js
@@ -1,6 +1,5 @@
 import * as PropTypes from 'prop-types'
 import React from 'react'
-import ReactDOM from 'react-dom'
 import { polyfill } from 'react-lifecycles-compat'
 
 import { timeoutsShape } from './utils/PropTypes'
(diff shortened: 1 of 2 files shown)


src/Transition.js
```diff
@@ -357,6 +361,8 @@ class Transition extends React.Component {
     delete childProps.onExiting
     delete childProps.onExited
 
+    childProps.ref = this.setChildNode
```
eps1lon: This will override the ref that is attached to child.

inomdzhon: And what about user component? Example
```
class Button extends React.Component {}
```

`ref` returned Symbol but not Element

bensampaio:
@inomdzhon @eps1lon what's keeping this MR from being merged?

inomdzhon:
> @inomdzhon @eps1lon what's keeping this MR from being merged?

@bensampaio, this PR [559](https://github.com/reactjs/react-transition-group/pull/559), no?

Issue #475, opened 2019-03-22
v2.7.0 breaks legacy module import


kaidjohnson:
> What is the current behavior?

```
var TransitionGroup = require('react-transition-group/TransitionGroup');
```
v2.7.0 returns an object `{ default: <TransitionGroup> }`

> What is the expected behavior?

v2.6.1 returns the `TransitionGroup` function directly.

`react-transition-group` is a ternary dependency from a few depths of our package tree, but I've confirmed that downgrading to 2.6.1 fixes the issue for us, but forces us to surface a ternary dependency in our package.json directly.

TrySound:
Do you mean transitive dependency?

TrySound:
@silvenon It's probably should be a major bump.

silvenon:
@TrySound I don't get it, do you know why this is happening and if it should? I thought CJS build stayed the same and that we just linked to it via a generated `package.json`.

silvenon:
@TrySound if this is expected and necessary behavior, then I'll revert the change in the next patch version and repeat the ESM commit as a breaking change.

silvenon:
We'll publish a new patch version within the next few days.

TrySound:
The problem is importing esm from cjs. Usually it's not a problem. But it is for example with material-ui.

TrySound:
@kaidjohnson Though there should be interop in that package. Which package depends on react-transition-group?

kaidjohnson:
@TrySound I believe I traced it to https://github.com/google-fabric/velocity-react/blob/master/src/velocity-transition-group.js#L56 v1.4.1 within Storybook v4.1.12

kaidjohnson:
:laughing: Yes, transitive :roll_eyes: Fridays and words don't always mix good.

TrySound:
Well, frontend projects written with commonjs is unexpected thing these days.

PR #488, opened 2019-04-15
Support esm via package.json routes

TrySound:
Ref #77

Upgraded semantic-release-alt-publish-dir to get correct module field
processing

https://github.com/jquense/semantic-release-alt-publish-dir/commit/795af9dae9a30d06268ac64a15e5072ea8d73906


TrySound:
Same PR as previously but with fixed module field
/cc @jquense @silvenon

taion:
same as https://github.com/reactjs/react-transition-group/pull/455

this is fine; we'll need to note the breaking change for semver

jquense:
what is the breaking change again?

TrySound:
Here's a breaking change
https://github.com/reactjs/react-transition-group/issues/475

taion:
@jquense it's https://github.com/reactjs/react-transition-group/issues/475#issuecomment-475782739

if you have a CJS module that imports the ESM build due to the inter-op, then you will actually get the ESM module this time around, which means you'll need to add the `.default`

PR #496, opened 2019-04-26
feat: Use fastdom to batch writes and avoid sync layout (take two)

diagramatics:
Continuing from #385. Sorry for the hijack @STRML I was going to force push to the PR branch instead but I realised it's on a separate repo 😢 

> When animating a large number of elements, it can be helpful to batch these animations up into a single mutate operation using `fastdom` to improve forced synchronous layout stalls.

This PR fixes the enter animations by running the class mutation for `enter` on the event handler, skipping fastdom and only relying it on the `-active` class. The difference is on instead of recalculating the DOM by forcefully invalidating it on every animation, the recalculation happens at the end of the frame per the browser logic. This is huge for animations where more than one element is being animated at the same frame. Results are on the flamecharts below on the `TransitionGroup` story when set to animate 3 elements at once.

Before:

image 
Notice the expensive style recalculation and relayouts on every animation done.

After:
image

PR also fixes the missing `webpack-atoms` dependency to run Storybook.

### DIFF
--- a/package.json
+++ b/package.json
@@ -59,9 +59,10 @@
   },
   "dependencies": {
     "dom-helpers": "^3.4.0",
+    "fastdom": "^1.0.8",
     "loose-envify": "^1.4.0",
     "prop-types": "^15.6.2",
-    "fastdom": "^1.0.8"
+    "webpack-atoms": "^10.0.2"
   },
   "devDependencies": {
     "@babel/cli": "^7.4.3",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -59,9 +59,10 @@
   },
   "dependencies": {
     "dom-helpers": "^3.4.0",
+    "fastdom": "^1.0.8",
     "loose-envify": "^1.4.0",
     "prop-types": "^15.6.2",
-    "fastdom": "^1.0.8"
+    "webpack-atoms": "^10.0.2"
```
STRML: Intended?

diagramatics:
Closing since #385 is updated now.

PR #497, opened 2019-04-30
chore(deps): add webpack-atoms to devDependencies

diagramatics:
Running `yarn storybook` throws an error that `webpack-atoms` isn't found. This PR fixes it.


jquense:
thanks!

PR #498, opened 2019-05-07
Documentation-fix for the child-function state

jdelStrother:
The child-function is never called with a state of 'unmounted', since `render()` bails early in that case.


taion:
Is that behavior correct, though?

jdelStrother:
I would have thought so.  You wouldn't want the child-function to be called after unmount, right?

silvenon:
Yep, this is true, just checked. Thanks!

silvenon:
Deployed docs with the patch. ✅

PR #499, opened 2019-05-09
fix" issue with dynamically applied classes not being properly removed for reentering items

Andarist:
Issue - https://codesandbox.io/s/zzxzjjwwmp

Notice how the reentered item (foo) has "exit" classes applied incorrectly.

### DIFF
--- a/src/CSSTransition.js
+++ b/src/CSSTransition.js
@@ -112,7 +112,7 @@ class CSSTransition extends React.Component {
   onExit = (node) => {
     this.removeClasses(node, 'appear');
     this.removeClasses(node, 'enter');
-    this.addClass(node, 'exit','base')
+    this.addClass(node, 'exit', 'base')
 
     if (this.props.onExit) {
       this.props.onExit(node)
--- a/src/CSSTransition.js
+++ b/src/CSSTransition.js
@@ -182,6 +182,8 @@ class CSSTransition extends React.Component {
       done: doneClassName
     } = this.appliedClasses[type]
 
+    this.appliedClasses[type] = {};
+
     if (baseClassName) {
       removeClass(node, baseClassName);
     }
--- a/src/CSSTransition.js
+++ b/src/CSSTransition.js
@@ -74,11 +74,16 @@ class CSSTransition extends React.Component {
   static defaultProps = {
     classNames: ''
   }
-  onEnter = (node, appearing) => {
-    const { className } = this.getClassNames(appearing ? 'appear' : 'enter')
 
+  appliedClasses = {
+    appear: {},
+    enter: {},
+    exit: {},
+  }
+
+  onEnter = (node, appearing) => {
     this.removeClasses(node, 'exit');
-    addClass(node, className)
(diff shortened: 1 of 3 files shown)


src/CSSTransition.js
```diff
@@ -150,44 +141,62 @@ class CSSTransition extends React.Component {
...
     };
   }
 
-  removeClasses(node, type) {
-    const { className, activeClassName, doneClassName } = this.getClassNames(type)
-    className && removeClass(node, className);
-    activeClassName && removeClass(node, activeClassName);
-    doneClassName && removeClass(node, doneClassName);
-  }
+  addClass(node, type, phase = 'base') {
```
taion:
```suggestion
  addClass(node, type) {
```

given how you're using this, i think it's a little cleaner to always pass in the `phase` explicitly. the default doesn't really feel "special" to me; it's just one of three possible phases, so i think it will be a little easier to read the code if we're always specifying the phase explicitly

src/CSSTransition.js
```diff
@@ -150,44 +141,62 @@ class CSSTransition extends React.Component {
     const isStringClassNames = typeof classNames === 'string';
     const prefix = isStringClassNames && classNames ? classNames + '-' : '';
```
taion:
```suggestion
    const prefix = isStringClassNames && classNames ? `${classNames}-` : '';
```

just making some un-related cleanups (:

src/CSSTransition.js
```diff
@@ -150,44 +141,62 @@ class CSSTransition extends React.Component {
...
 
-    let className = isStringClassNames ?
+    let baseClassName = isStringClassNames ?
       prefix + type : classNames[type]
```
taion:
```suggestion
      `${prefix}${type}` : classNames[type]
```

src/CSSTransition.js
```diff
@@ -150,44 +141,62 @@ class CSSTransition extends React.Component {
...
 
     let activeClassName = isStringClassNames ?
-      className + '-active' : classNames[type + 'Active'];
+      baseClassName + '-active' : classNames[type + 'Active'];
```
taion:
```suggestion
      `${baseClassName}-active` : classNames[`${type}Active`];
```

src/CSSTransition.js
```diff
@@ -150,44 +141,62 @@ class CSSTransition extends React.Component {
...
 
     let doneClassName = isStringClassNames ?
-      className + '-done' : classNames[type + 'Done'];
+      baseClassName + '-done' : classNames[type + 'Done'];
```
taion:
```suggestion
      `${baseClassName}-done` : classNames[`${type}Done`];
```

Andarist:
@taion thanks for the review! I've applied all your suggested code changes

Andarist:
No pressure - just a question. Are there chances for a new release with this soon(-ish) or should we use a fork temporarily?

jquense:
aaarrrrg i messed up the commit name. it'll get published once i fix it

Andarist:
Thanks for the quick release ❤️

Issue #502, opened 2019-05-12
undefined being added to the "class" attribute


satazor:
> What is the current behavior?

When `appear` is enabled and `appearDone` is not defined in the class names, `undefined` is added.

> What is the expected behavior?

`appearDone` is optional, therefore `undefined` shouldn't be added to the "class" attribute

> Could you provide a [CodeSandbox](https://codesandbox.io/) demo reproducing the bug?

I feel the bug is pretty straightforward to fix, so I feel that creating a demo would not be necessary. The issue is in this line: https://github.com/reactjs/react-transition-group/blob/db834960c194ff742b02b771d02151896a159208/src/CSSTransition.js#L169

It should check if `this.getClassNames('enter').doneClassName` is a string before appending it.

PR #504, opened 2019-05-20
Fix adding undefined to class

mmarkelov:
Fix https://github.com/reactjs/react-transition-group/issues/502

### DIFF
--- a/src/CSSTransition.js
+++ b/src/CSSTransition.js
@@ -164,9 +164,10 @@ class CSSTransition extends React.Component {
 
   addClass(node, type, phase) {
     let className = this.getClassNames(type)[`${phase}ClassName`];
+    const { doneClassName } = this.getClassNames('enter');
 
-    if (type === 'appear' && phase === 'done') {
-      className += ` ${this.getClassNames('enter').doneClassName}`;
+    if (type === 'appear' && phase === 'done' && doneClassName) {
+      className += ` ${doneClassName}`;
     }
 
     // This is for to force a repaint,
@@ -176,8 +177,10 @@ class CSSTransition extends React.Component {
       node && node.scrollTop;
     }
 
-    this.appliedClasses[type][phase] = className
-    addClass(node, className)
+    if (className) {
+      this.appliedClasses[type][phase] = className
+      addClass(node, className)
+    }
   }
 
   removeClasses(node, type) {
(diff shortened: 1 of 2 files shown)


src/CSSTransition.js
```diff
@@ -176,8 +177,10 @@ class CSSTransition extends React.Component {
       node && node.scrollTop;
     }
 
-    this.appliedClasses[type][phase] = className
-    addClass(node, className)
+    if (className) {
+      this.appliedClasses[type][phase] = className
```
taion: this should be outside the conditional, no? we want to e.g. reset old values

mmarkelov: @taion I think that it is unnecessary to add falsy value to object and run **addClass** function, i can remove this conditional if you think that it will be better approach

taion: well, we want to clear any previous value that was there, no? e.g. if this already ran a transition, and the class mappings are dynamic, then we don't want to keep around an old stale value from the last time the transition ran?

mmarkelov: **className** will be **undefined** for this case (when **doneClassName** is **undefined**), so if it is **undefined** we could pass through: 
```javascript
this.appliedClasses[type][phase] = className
```
**appliedClasses** is used in **removeClasses** function, and it also check for **undefined**

also we can pass through:
```javascript
addClass(node, className)
```
cause it check if **className** is falsy.
So I suppose it is equivalent, but with this condition, we can skip  **addClass** and pass **undefined** value to **appliedClasses**

taion: i guess we wipe it in https://github.com/reactjs/react-transition-group/blob/5e2c1c4cb300b59d1c1242cb15b9607fc9caf723/src/CSSTransition.js#L193 anyway

src/CSSTransition.js
```diff
@@ -198,6 +201,10 @@ class CSSTransition extends React.Component {
     if (doneClassName) {
       removeClass(node, doneClassName);
     }
+    // Remove class atribute if it is empty
```
taion: do we need this change? this seems like essentially a separate thing. does it hurt to leave an empty `class` attribute around? we wouldn't know if e.g. the node originally had a `class=""`, would we?

mmarkelov: @taion youp. I don't know if it is right approach. I add it to cleanup for this specific case, cause it just add class to div, like so:
```html
<div class></div>
<div class></div>
<div class></div>
``` 
Maybe there is another way to fix this?
It doesn't hurt to add `class` attribute, so I can delete this part

taion: I think it's harmless.

gpbl:
Pinging @jquense – would be nice to get this merged :) Thanks!

PR #506, opened 2019-05-30
feat: add global transition disable switch

jquense:


www/src/pages/testing.js
```diff
@@ -0,0 +1,68 @@
+import { graphql } from 'gatsby';
+import PropTypes from 'prop-types';
+import React from 'react';
+import { Container } from 'react-bootstrap';
+import Layout from '../components/Layout';
+import Example from '../components/Example';
+
+const propTypes = {
+  location: PropTypes.object.isRequired,
+  data: PropTypes.shape({
```
taion: why do we need all this 🤨

jquense: we don't i just copy/pasted from the other page

piotr-cz:
IMHO docs should be updated with [info](https://github.com/reactjs/react-transition-group/blob/v4.1.0/www/src/pages/testing.js#L28-L35) how to use this feature

PR #507, opened 2019-06-04
fix: adds missing dependency @babel/runtime

marco-silva0000:
Using yarn with pnp enabled enforces a correct dependency structure. While using any module that has this package as a dependency I get the following error:

```
ERROR in /home/msilva/.cache/yarn/v4/npm-react-transition-group-4.1.0-7b50c0a93a6c127336187252c3c1a70eff3304ce/node_modules/react-transition-group/esm/CSSTransition.js
Module not found: Error: Package "react-transition-group@4.1.0" (via "/home/msilva/.cache/yarn/v4/npm-react-transition-group-4.1.0-7b50c0a93a6c127336187252c3c1a70eff3304ce/node_modules/react-transition-group/esm/CSSTransition.js") is trying to require the package "@babel/runtime" (via "@babel/runtime/helpers/esm/extends") without it being listed in its dependencies (react, react-dom, dom-helpers, loose-envify, prop-types, react-transition-group)
 @ /home/msilva/.cache/yarn/v4/npm-react-transition-group-4.1.0-7b50c0a93a6c127336187252c3c1a70eff3304ce/node_modules/react-transition-group/esm/CSSTransition.js 1:0-58 228:43-51
 @ /home/msilva/.cache/yarn/v4/npm-react-transition-group-4.1.0-7b50c0a93a6c127336187252c3c1a70eff3304ce/node_modules/react-transition-group/esm/index.js
 @ /home/msilva/.cache/yarn/v4/npm-@material-ui-core-4.0.2-c7a07c1d1b13b94adf479d7267b51cb682fcb279/node_modules/@material-ui/core/Slide/Slide.js
```

to fix this I added the latest version of @babel/runtime to the dependencies.

### DIFF
--- a/package.json
+++ b/package.json
@@ -58,7 +58,7 @@
     "react-dom": ">=16.6.0"
   },
   "dependencies": {
-    "@babel/runtime": "^7.4.5"
+    "@babel/runtime": "^7.4.5",
     "dom-helpers": "^3.4.0",
     "loose-envify": "^1.4.0",
     "prop-types": "^15.6.2"
--- a/package.json
+++ b/package.json
@@ -58,6 +58,7 @@
     "react-dom": ">=16.6.0"
   },
   "dependencies": {
+    "@babel/runtime": "^7.4.5"
     "dom-helpers": "^3.4.0",
     "loose-envify": "^1.4.0",
     "prop-types": "^15.6.2"


taion:
I believe this is a bug with our build process right now. The ESM build uses Babel runtime, while the CJS build does not. Can we sync them up?

TrySound:
Not necessary. Missing dependency is a bug. Need just to add it.

taion:
The issue is that I don't think we're intentionally applying the runtime transform here. We don't do so for the CJS build, so why should we do so for the ESM build?

TrySound:
To not duplicate babel helpers for each module in this package. ESM thing is used for browsers. Node thing is less important. But we may do that too.

taion:
Ah, I see that the config for the ESM build doesn't exclude runtime. I guess that must be intentional.

PR #512, opened 2019-06-18
docs: Add anchor for props section

eps1lon:
Adds an anchor to the props section in addition to the existing anchors for each individual prop.


jquense:
thanks!

PR #514, opened 2019-06-27
Added alternative to findDOMNode

AndrewMusgrave:
This approach provides an alternative to `ReactDOM.findDOMNode` for those who would like to enable `React.StrictMode` without console errors from `react-transition-group`.

### DIFF
--- a/test/Transition-test.js
+++ b/test/Transition-test.js
@@ -469,4 +469,30 @@ describe('Transition', () => {
       wrapper.setState({ in: false })
     })
   })
+
+  describe('transition node', () => {
+    it('delegates to findDOMNode when a transntionNode is not provided', () => {
+      const findDOMNodeSpy = jest.spyOn(ReactDOM, 'findDOMNode');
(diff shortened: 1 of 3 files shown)


beefchimi:
@silvenon any chance this PR can get reviewed?

test/Transition-test.js
```diff
@@ -469,4 +469,30 @@ describe('Transition', () => {
...
+        </Transition>
+      )
+
+      expect(findDOMNodeSpy).toHaveBeenCalledTimes(1);
+      findDOMNodeSpy.mockRestore();
+    })
+
+    it('uses transitionNode when provided', () => {
+      const onEnterSpy = jest.fn();
+      const transitionNode = document.createElement('span');
```
eps1lon: But that is not the element that is transitioning. It should be the child of `Transition` which is a `div` in this case.

JSteunou: should use ref

PR #516, opened 2019-07-02
Ensure SwitchTransition component is exported

chrisbrown-io:
**Issue:**
New `SwitchTransition` component isn't exported from `src/index.js` and cannot be used unless imported directly.

**Fix:**
Added export of `SwitchTransition` from `src/index.js`. Also updated export of `SwitchTransition` to be default export, with other exports being named exports, following the convention of the other components.

### DIFF
--- a/src/SwitchTransition.js
+++ b/src/SwitchTransition.js
@@ -96,7 +96,7 @@ const enterRenders = {
  * }
  * ```
  */
-export class SwitchTransition extends React.Component {
+class SwitchTransition extends React.Component {
   state = {
     status: ENTERED,
     current: null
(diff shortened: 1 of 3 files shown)


kumar303:
Could you describe the symptoms of this bug before fixing it? We're trying to track down a failure in our app possibly related to react-transition-group 4.2.0. We see `ChunkLoadError: Loading chunk N failed`.

jquense:
The only symptom you should have seen between 4.2.0 and 4.2.1 would be if you tried to import SwitchTransition, which isn't exported correctly in `4.2.0`. If ya'll running into trouble i'd check that 4.2.0 exactly vs the previous version is causing the error. That said `4.2.0` only added a new component it really shouldn't be causing an issue in existing code!

kumar303:
ok, thanks. Just to follow up, we are less sure that the error we're seeing is related to our upgrade of react-transition-group. It's possible that the transition component is just the first one to trigger a render of the error'ing code path.

PR #518, opened 2019-07-07
Add codesandbox example for switch-transition

dimensi:
https://codesandbox.io/s/50p5wr3jwl


silvenon:
Thanks for the suggestion 😃 I built my own demo based on yours, you can see it in [`SwitchTransition` docs](http://reactcommunity.org/react-transition-group/switch-transition).

PR #525, opened 2019-07-23
fix(readme): add guidance on TypeScript in README

Setsun:
## Description
Adds a short section about TypeScript to the README.md, to provide guidance to users who may be using TypeScript in their projects.

### To-do
There could be a section added about `flow-typed`, though I am unfamiliar with that particular workflow.

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,6 +13,13 @@ A set of components for managing component states (including mounting and unmoun
 - [**Main documentation**](https://reactcommunity.org/react-transition-group/)
 - [Migration guide from v1](/Migration.md)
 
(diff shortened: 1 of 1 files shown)


arshdkhn1:
It would be better to add what version of @types/react-transition-group, one should install. When install `react-transition-group@1.x` then even by installing types by `npm install @types/react-transition-group`, TS will complain about declaration file. 

I got past around this issue by installing `npm install @types/react-transition-group@1.x`. Not sure that's the right way but it solves the problem on my side.

jquense:
thanks!

PR #526, opened 2019-07-25
SwitchTransition children may be equal if both keys are null

AdrianFleming:
If the SwitchTransition's children both have a null key then they were not equal. If the Router's location has no key then the children were being continually re-rendered with the animation applied.

### DIFF
--- a/src/SwitchTransition.js
+++ b/src/SwitchTransition.js
@@ -8,8 +8,7 @@ function areChildrenDifferent(oldChildren, newChildren) {
   if (
     React.isValidElement(oldChildren) &&
     React.isValidElement(newChildren) &&
-    oldChildren.key != null &&
-    oldChildren.key === newChildren.key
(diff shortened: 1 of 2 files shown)


src/SwitchTransition.js
```diff
@@ -8,8 +8,7 @@ function areChildrenDifferent(oldChildren, newChildren) {
   if (
     React.isValidElement(oldChildren) &&
     React.isValidElement(newChildren) &&
-    oldChildren.key != null &&
-    oldChildren.key === newChildren.key
+    ((oldChildren.key == null && newChildren.key == null) || oldChildren.key === newChildren.key)
```
taion: Why would we want to treat `null` and `undefined` the same here?

AdrianFleming: @taion, OK. This is how I initially coded it but decided it would be better handled in the component. I'll close the fork therefore.

PR #537, opened 2019-08-12
Remove warning about deprecated findDOMNode usage

iegik:
https://reactjs.org/docs/strict-mode.html#warning-about-deprecated-finddomnode-usage

https://github.com/reactstrap/reactstrap/issues/1159

### DIFF
--- a/src/Transition.js
+++ b/src/Transition.js
@@ -117,6 +117,7 @@ class Transition extends React.Component {
     let initialStatus
 
     this.appearStatus = null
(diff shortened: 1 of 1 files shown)


jquense:
Hi there, thanks for the PR we appreciate taking the time to try and fix something. Unfortunately it looks like your fix is incomplete and nonfunctional. Refs don't really work like this and in order to remove findDOMNode we need to make broader sweeping changes to the API

PR #540, opened 2019-08-18
fix: appearing animation with Transition

### LINKED ISSUE #538: Transition Component with "appear={true}" props not transition at the first appearing

Transition Component with `appear={true}` props not transition at the first appearing.

the content of console.log:
```
state: exited 
onEnter: appearing... 
onEnter: entering... 
state: entering 
state: entered 
```
the text transition with CSS-styles should be move from left to right at the first appearing, according to [Transition-prop-appear](http://reactcommunity.org/react-transition-group/transition#Transition-prop-appear)

> Could you provide a CodeSandbox demo reproducing the bug?

Yes, here we go:
[transitiongroup-component-yo5u3](https://codesandbox.io/s/transitiongroup-component-yo5u3)

koba04:
This fixes #538.
Currently, changing `appearing` status is synchronously on cDM, so transitions never been applied.
The following is a minimum case for that.

https://codepen.io/koba04/pen/JjPKmMw

So I've added a reflow before the transition as well as CSSTransition.

### DIFF
--- a/src/Transition.js
+++ b/src/Transition.js
@@ -167,9 +167,11 @@ class Transition extends React.Component {
   // }
 
   componentDidMount() {
-    const node = ReactDOM.findDOMNode(this)
     // This is necessary to make a transition happen
-    reflow(node)
+    if (this.appearStatus === ENTERING) {
+      const node = ReactDOM.findDOMNode(this)
+      reflow(node)
+    }
     this.updateStatus(true, this.appearStatus)
   }
 
--- a/src/Transition.js
+++ b/src/Transition.js
@@ -190,6 +190,12 @@ class Transition extends React.Component {
         }
       }
     }
+    // if mountOnEnter is true, this is the time that children is mounted.
+    // so we have to trigger reflow the same as cDM
+    if (this.props.mountOnEnter && nextStatus === ENTERING) {
+      const node = ReactDOM.findDOMNode(this)
+      reflow(node)
+    }
     this.updateStatus(false, nextStatus)
   }
 
--- a/src/Transition.js
+++ b/src/Transition.js
@@ -4,6 +4,7 @@ import ReactDOM from 'react-dom'
 
 import config from './config'
 import { timeoutsShape } from './utils/PropTypes'
+import { reflow } from './utils/dom'
 import TransitionGroupContext from './TransitionGroupContext'
 
 export const UNMOUNTED = 'unmounted'
@@ -166,6 +167,9 @@ class Transition extends React.Component {
   // }
 
   componentDidMount() {
+    const node = ReactDOM.findDOMNode(this)
+    // This is necessary to make a transition happen
+    reflow(node)
     this.updateStatus(true, this.appearStatus)
   }
 
--- a/src/CSSTransition.js
+++ b/src/CSSTransition.js
@@ -6,6 +6,7 @@ import React from 'react';
 
 import Transition from './Transition';
 import { classNamesShape } from './utils/PropTypes';
+import { reflow } from './utils/dom';
 
 const addClass = (node, classes) => node && classes && classes.split(' ').forEach(c => addOneClass(node, c));
 const removeClass = (node, classes) => node && classes && classes.split(' ').forEach(c => removeOneClass(node, c));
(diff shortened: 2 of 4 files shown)


koba04:
I've updated a way to fix this and updated the PR description.

src/Transition.js
```diff
@@ -166,6 +167,9 @@ class Transition extends React.Component {
   // }
 
   componentDidMount() {
+    const node = ReactDOM.findDOMNode(this)
+    // This is necessary to make a transition happen
+    reflow(node)
```
taion: I thought we didn't want to unconditionally do this for all transitions? cc @jquense

jquense: we don't i don't think

koba04: That makes sense so I've updated this PR to trigger reflow only that the component has an appearing animation.

lifeiscontent:
@koba04 any updates on this? I'm currently running into this issue.

koba04:
@lifeiscontent I think this is in review status.

lifeiscontent:
@koba04 I'm seeing an issue in the latest released version of react-transition-group where the states are not being applied properly when `mountOnEnter` and `unmountOnExit` are used together.

koba04:
@lifeiscontent Could you create a reproduce case for that?

lifeiscontent:
@koba04 sure, when I have some free time I’ll create a codesandbox :)

lifeiscontent:
@koba04 here you go: https://codesandbox.io/s/hello-world-nvjwn I left a comment in the example of what to comment out to see expected behavior.

lifeiscontent:
@koba04 any updates on a fix for that?

koba04:
@lifeiscontent
I've fixed the case `mountOnEnter` is `true` at https://github.com/reactjs/react-transition-group/pull/540/commits/9d94aaec071d6ad6f2eeb134cfbb7333c6ffa8cd.
Thanks!

lifeiscontent:
@koba04 @jquense any idea when this might get merged/released?

lifeiscontent:
@koba04 ping

koba04:
@lifeiscontent I think this is waiting a review.

lifeiscontent:
@jquense @taion can we get this reviewed? :)

lifeiscontent:
@koba04 how long does a review take?

koba04:
@lifeiscontent I have no idea. I don't have anything I can do.

lifeiscontent:
ping @taion @jquense

lifeiscontent:
ping @silvenon @gaearon @Andarist

jtomaszewski:
If you're not sure of the risk of that change, maybe we can at least add those reflows behind some boolean prop (`<Transition autoReflow>`) ?

I wanted to use `Transition` component today in my repo that uses styled-components, couldn't get it done for most of the day; later on I used CSSTransition and it worked straight away... because reflows are included in `CSSTransition` but not in `Transition`. I've been using this library for more than a year now, (I love it! It's so simple and is enough for 99% of animation cases), but still keep forgetting about it.

lifeiscontent:
@silvenon do you have time to review this please :)

silvenon:
I'm sorry that you spent effort on this, but the documentation for `Transition` explains:

`Transition` is a platform-agnostic base component. If you're using transitions in CSS, you'll probably want to use [`CSSTransition`](http://reactcommunity.org/react-transition-group/css-transition) instead. It inherits all the features of `Transition`, but contains additional features necessary to play nice with CSS transitions (hence the name of the component).

`Transition` is a primitive intended for alternative means of transitioning. Forcing a reflow is the key feature of `CSSTransition` intended precisely for CSS transitions, which you're using in your demo, and considering that this additional clarification was already in place before your PR, is there any specific reason why you're insisting on using `Transition`?

lifeiscontent:
@silvenon even swapping out the example to use CSSTransition doesn't work as expected: https://codesandbox.io/s/hello-world-nvjwn?file=/pages/index.js

silvenon:
Commenting out suggested props looks the same to me.

In any case, you should use classes applied by `CSSTransition` to achieve the desired result, please read [its documentation](http://reactcommunity.org/react-transition-group/css-transition).

lifeiscontent:
@silvenon I've tried everything you've suggested since its a `nextJS` project you might of not been rendering the updated code as its a server side project it takes a little finagling to get it to work on codesandbox. The issue is that the exit transition works (this is fine in all of the case you've suggested) but during the appear animation it never calls the entering animation state because the component mounts and skips the animation entirely, (this is due to react rendering on the server) (and is also the bug that I wanted to have fixed)

silvenon:
Your demo is missing the `appear` prop, which tells `Transition` and `CSSTransition` to run the enter transition on mount if `in={true}`, in that case you need to style `*-appear` and `*-appear-active` classes. Refresh the preview in the following CodeSandbox to see the effect:

[Edit magical-cdn-of8x2](https://codesandbox.io/s/magical-cdn-of8x2?fontsize=14&hidenavigation=1&theme=dark)

If you find a bug in this feature, open a new issue with either a CodeSandbox or a repo which reproduces the problem.

PR #552, opened 2019-09-10
import style dom-helper from 'dom-helpers/css' instead of 'dom-helpers/style' 

flurmbo:
Currently when running `npm run storybook` I get this compilation error:
```
ERROR in ./stories/transitions/Bootstrap.js
Module not found: Error: Can't resolve 'dom-helpers/style'
```

This is fixed by changing the import statement to import from `dom-helpers/css` instead of `dom-helpers/style`.

### DIFF
--- a/stories/transitions/Bootstrap.js
+++ b/stories/transitions/Bootstrap.js
@@ -1,5 +1,5 @@
 import React from 'react';
-import style from 'dom-helpers/style';
+import style from 'dom-helpers/css';
 
 import Transition, { EXITED, ENTERED, ENTERING, EXITING }
   from '../../src/Transition';


stories/transitions/Bootstrap.js
```diff
@@ -1,5 +1,5 @@
 import React from 'react';
-import style from 'dom-helpers/style';
+import style from 'dom-helpers/css';
```
taion: we should import this as `css`, then?

flurmbo: Hi @taion, I'm not sure. The dom-helpers package exports something called `style` from its index.js file, so that's what you would need to import if you were using the whole package instead of just the one function. But that function is being imported from `css.ts`, which is what causes this issue here. I opened an issue on the dom-helpers repo since it's unclear what the API is supposed to look like.

taion: @jquense ?

jquense: it's just a naming oddity, I don't  really care which one we go with. `css` is more jquery-y but not really correct (and conflicts with styled-component type libraries)

PR #564, opened 2019-10-23
prefix UNSAFE_ to componentWillReceiveProps lifecycle method in tests

AlbertCarreras:
Unsafe lifecycles  (componentWillReceiveProps) will continue to work until upcoming React version 17. 
[docs](https://reactjs.org/docs/react-component.html#unsafe_componentwillreceiveprops)

Gradual migration for these lifecycle methods includes prefixing `UNSAFE_` which will prevent deprecation errors and indicate that using these `UNSAFE_` lifecycles will be more prone to bugs in the future.

### DIFF
--- a/test/TransitionGroup-test.js
+++ b/test/TransitionGroup-test.js
@@ -102,7 +102,7 @@ describe('TransitionGroup', () => {
 
   it('should not throw when enter callback is called and is now leaving', () => {
     class Child extends React.Component {
-      componentWillReceiveProps() {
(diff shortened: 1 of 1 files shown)


pelotom:
It seems like it would be better to actually fix the problem rather than squashing the warnings?

taion:
Why does it matter for the tests here? They're still running against React 16. The code itself is React 17 compatible.

iamandrewluca:
@taion @silvenon This PR is outdated, the test was deleted in #611

PR #567, opened 2019-10-29
Prefix more lifecycles with UNSAFE_ (fix Warnings)

devinrhode2:
This is a naive patch. It's working with react 16.11.0. I had thought I was using react 16.10. Whadayaknow.

### DIFF
--- a/index.js
+++ b/index.js
@@ -454,7 +454,7 @@ var TransitionGroup = function (_React$Component) {
     return _this;
   }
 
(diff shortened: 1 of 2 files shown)


koba04:
`react-transition-group` is no longer using deprecated lifecycle methods and `react-transition-group` now supports React v17.  https://github.com/reactjs/react-transition-group/pull/317
Thank you for your pull request!

PR #585, opened 2020-01-17
Update SwitchTransition Example

lfarrel6:
The example for SwitchTransition isn't a complete example, the FadeTransition element isn't defined anywhere which takes from the explanation of SwitchTransition.

I changed out the FadeTransition for CSSTransition and defined the fade classes to remove any unnecessary confusion.


silvenon:
Yep, that's much better, thaks! ❤️

Issue #341, opened 2018-05-21
CSSTransition with transitions disabled end up with both -exit-done and -enter-done classes


goffioul:
**Do you want to request a *feature* or report a *bug*?**
Bug

**What is the current behavior?**
When defining a CSSTransition component with both "enter" and "exit" set to false, the component ends up with both -exit-done and -enter-done classes after 2 transitions.

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem via https://jsfiddle.net or similar (template: https://jsfiddle.net/m4mb9Lg3/4/).**
https://jsfiddle.net/5sLyLhyg/1/

**What is the expected behavior?**
Only one of the -exit-done and -enter-done should exist on the DOM node.

**Which versions, and which browser / OS are affected by this issue? Did this work in previous versions?**
Tested on Chrome 66.

PR #590, opened 2020-02-25
Remove enter and exit classes if config.disabled is true

karlwills:
This pull request relates to https://github.com/reactjs/react-transition-group/issues/341 and ensures that only the `*-enter-done` and `*-exit-done` should exist on the tested DOM node when `config.disabled` is set to`true`

Credit goes to @joefurniss for spotting this issue.

#### Testing Components with Transitions
https://reactcommunity.org/react-transition-group/testing/

### DIFF
--- a/src/CSSTransition.js
+++ b/src/CSSTransition.js
@@ -4,6 +4,8 @@ import addOneClass from 'dom-helpers/addClass';
 import removeOneClass from 'dom-helpers/removeClass';
 import React from 'react';
 
+import config from './config';
+
 import Transition from './Transition';
 import { classNamesShape } from './utils/PropTypes';
 
@@ -108,6 +110,10 @@ class CSSTransition extends React.Component {
     this.removeClasses(node, type);
     this.addClass(node, type, 'done');
 
+    if(this.config.disabled) {
+      this.removeClasses(node, 'exit');
+    }
+
     if (this.props.onEntered) {
       this.props.onEntered(node, appearing)
     }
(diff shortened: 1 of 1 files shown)


taion:
This doesn't actually fix #341, though, does it? That one doesn't just cover cases with `config.disabled`?

karlwills:
@taion No it doesn't fix that issue, It does relate to it though.

However, it _does_ fix the issue I was having (and feel that should be fixed in the codebase) with both `*-enter-done` and `*-exit-done` classes appearing when testing components that rely on CSSTransition

src/CSSTransition.js
```diff
@@ -108,6 +110,10 @@ class CSSTransition extends React.Component {
     this.removeClasses(node, type);
     this.addClass(node, type, 'done');
 
+    if (config.disabled) {
+      this.removeClasses(node, 'exit');
```
taion:
```js
  onEntered = (node, appearing) => {
    this.removeClasses(node, 'exit');

    const type = appearing ? 'appear' : 'enter'
    this.removeClasses(node, type);
    this.addClass(node, type, 'done');

    if (this.props.onEntered) {
      this.props.onEntered(node, appearing)
    }
  }
```

src/CSSTransition.js
```diff
@@ -135,6 +141,10 @@ class CSSTransition extends React.Component {
     this.removeClasses(node, 'exit');
     this.addClass(node, 'exit', 'done');
 
+    if (config.disabled) {
+      this.removeClasses(node, 'enter');
```
taion:
```js
  onExited = (node) => {
    this.removeClasses(node, 'enter');
    this.removeClasses(node, 'appear');

    this.removeClasses(node, 'exit');
    this.addClass(node, 'exit', 'done');

    if (this.props.onExited) {
      this.props.onExited(node)
    }
  }
```

matt-oconnell:
@taion @karlwills We're running into this issue as well. Are there any blockers to @taion 's suggested fixes?

I can try and take over the PR if you guys need some more resources on this

PR #591, opened 2020-02-26
test: remove deprecated findDOMNode from tests

iamandrewluca:


jquense:
not sure i see why but thanks!

Issue #601, opened 2020-03-25
New mode for SwitchTransition - `in-out-at-once`


jtomaszewski:
Currently `SwitchTransition` has only [two modes](https://github.com/reactjs/react-transition-group/blob/cfd0070765d631112c5da817cef82742566d2ed0/src/SwitchTransition.js#L78-L79):

>  * If the `out-in` mode is selected, the `SwitchTransition` waits until the old child leaves and then inserts a new child.
>  * If the `in-out` mode is selected, the `SwitchTransition` inserts a new child first, waits for the new child to enter and then removes the old child

What I'd like to have, is: "insert a new child first, animate both old and new child at once, and then remove the old child".

What do you think? I could make a PR for that.

taion:
You don't need `<SwitchTransition>` for that. That's just the behavior you'd get if you used a normal `<TransitionGroup>`.

jtomaszewski:
Thanks, you're right. Somehow I haven't thought about this. Maybe I'll suggest some small PR with an update to the documentation instead.

PR #602, opened 2020-03-25
docs(SwitchTransition): Add info when TransitionGroup should be used instead

jtomaszewski:
See https://github.com/reactjs/react-transition-group/issues/601 for why this note might be useful.


taion:
cc @dimensi

dimensi:
Well, if it helps to understand the library better, I don't mind.

PR #611, opened 2020-04-17
chore: Bump deps

taion:


taion:
d'oh

taion:
@silvenon okay, the storybook was screwed up more broadly because the transitions themselves were disabled. i fixed it and modernized some of the code. i also took out those 2 stale test cases

iamandrewluca:
My suggestion would be to revert back `react` and `react-dom` `devdependencies` versions to

```
"react": "~16.6.3",
"react-dom": "~16.6.3",
```

because they are listed also in `peerDependencies`, and there is a risk to use some `react` APIs that are not supported by the dependencies of `react-transition-group` users

taion:
The peer dependencies use a `>=` range. I expect most users are actually on the latest React versions at this point, so it seems to me to make more sense to test against those.

iamandrewluca:
> I expect most users are actually on the latest React versions at this point

I also thought this on `redux-form` but it seems is not the case.

What I mean above is that for example I'm working on a feature `react-transition-group` and I will be tempted to use a react feature that was introduced in react `>16.8`. And we release this new feature. And some users of the library, may still have react `16.6` installed, and their build will fail.

silvenon:
I think it's best to set up a sanity test on CI with lowest supported React version.

PR #619, opened 2020-05-05
Support SSR

### LINKED ISSUE #618: SSR in 4.4.0 crashes

Running this module on node.js (for example during ssr) results in: `ReferenceError: Element is not defined`

Importing `react-transition-group` does not result in an error

The problem is this line:
https://github.com/reactjs/react-transition-group/blob/c8631587f9e1d18f5161f5490e6f5c225f06c6ae/src/Transition.js#L397

A [possible solution](https://stackoverflow.com/questions/48007326/what-is-the-correct-proptype-for-a-ref-in-react) is: `Element = typeof Element === 'undefined' ? function(){} : Element`

perrin4869:
Fixes #618 and adds regression tests to make sure the lib loads in node environment

### DIFF
--- a/src/Transition.js
+++ b/src/Transition.js
@@ -394,7 +394,11 @@ Transition.propTypes = {
    *     (see
    *     [test/CSSTransition-test.js](https://github.com/reactjs/react-transition-group/blob/13435f897b3ab71f6e19d724f145596f5910581c/test/CSSTransition-test.js#L362-L437)).
    */
-  nodeRef: PropTypes.shape({ current: PropTypes.instanceOf(typeof Element === 'undefined' ? function() {} : Element) }),
+  nodeRef: PropTypes.shape({ 
+    current: typeof Element === 'undefined' 
+      ? PropTypes.any
+      : PropTypes.instanceOf(Element) 
+  }),
 
   /**
    * A `function` child can be used instead of a React element. This function is
--- a/src/Transition.js
+++ b/src/Transition.js
@@ -394,10 +394,10 @@ Transition.propTypes = {
    *     (see
    *     [test/CSSTransition-test.js](https://github.com/reactjs/react-transition-group/blob/13435f897b3ab71f6e19d724f145596f5910581c/test/CSSTransition-test.js#L362-L437)).
    */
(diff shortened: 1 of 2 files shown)


src/Transition.js
```diff
@@ -394,7 +394,7 @@ Transition.propTypes = {
    *     (see
    *     [test/CSSTransition-test.js](https://github.com/reactjs/react-transition-group/blob/13435f897b3ab71f6e19d724f145596f5910581c/test/CSSTransition-test.js#L362-L437)).
    */
-  nodeRef: PropTypes.shape({ current: PropTypes.instanceOf(Element) }),
+  nodeRef: PropTypes.shape({ current: PropTypes.instanceOf(typeof Element === 'undefined' ? function() {} : Element) }),
```
oliviertassinari, line 397: Better not to rely on Element at all to support different windows http://tobyho.com/2011/01/28/checking-types-in-javascript/.

jquense:
```suggestion
  nodeRef: PropTypes.shape({ 
    current: typeof Element === 'undefined' 
      ? PropTypes.any
      : PropTypes.instanceOf(Element) 
   }),
```

src/Transition.js
```diff
@@ -394,7 +394,11 @@ Transition.propTypes = {
    *     (see
    *     [test/CSSTransition-test.js](https://github.com/reactjs/react-transition-group/blob/13435f897b3ab71f6e19d724f145596f5910581c/test/CSSTransition-test.js#L362-L437)).
    */
-  nodeRef: PropTypes.shape({ current: PropTypes.instanceOf(Element) }),
+  nodeRef: PropTypes.shape({ 
+    current: typeof Element === 'undefined' 
+      ? PropTypes.any
+      : PropTypes.instanceOf(Element)
```
oliviertassinari, line 400: Could we not rely on "Element" for this prop-type? My concern is equivalent to https://github.com/merceyz/typescript-to-proptypes/issues/3.

oliviertassinari: The problem can be noticed on https://github.com/mui-org/material-ui/issues/21858#issuecomment-661173090.

PR #620, opened 2020-05-05
CI: Fix ESLint errors in www

silvenon:


.travis.yml
```diff
@@ -3,6 +3,8 @@ language: node_js
 node_js:
   - "11.10"
 cache: yarn
+install:
+  - yarn install --cwd www
```
taion, line 7: i think this needs to be

```
  - yarn
  - yarn --cwd www
```

otherwise you don't do the install in the root?

silvenon: You're right! I pushed the change.

PR #621, opened 2020-05-05
refactor(Transition): simplify render function

iamandrewluca:


src/Transition.js
```diff
@@ -341,44 +341,38 @@ class Transition extends React.Component {
...
+      onExit: _onExit,
+      onExiting: _onExiting,
+      onExited: _onExited,
+      nodeRef: _nodeRef,
+      ...childProps
+    } = this.props
+
+    const child = typeof children === 'function'
+      ? children(status, childProps)
+      : React.cloneElement(React.Children.only(children), childProps)
```
taion, line 372: **suggestion:** inline this as something like:

```js
return (
  <TransitionGroupContext.Provider value={null}>
    {typeof children === 'function' ? ...}
  </TransitionGroupContext.Provider>
);
```

PR #629, opened 2020-05-20
refactor(Transition): improve `timeout` prop as a object and make optional

iamandrewluca:
Purpose of #464 was to make timeout optional and default to 0
if `addEndListener` prop is missing, but propType was still marked as
required

Also removed unused class `SimpleSet`


src/Transition.js
```diff
@@ -590,3 +575,17 @@ Transition.ENTERED = ENTERED
 Transition.EXITING = EXITING
 
 export default Transition
+
+function getTimeouts(timeout) {
```
taion, line 579: **nitpick:** we should generally define methods before we use them, so maybe lift this to the top?

taion, line 579: **suggestion:** it would probably be a bit cleaner to refactor this to also take the _type_ of timeout desired – e.g. there's no need to calculate the "exit" timeout in the "enter" flow, and no need to calculate the "enter/appear" timeout in the "exit" flow

just would be nice to clean up APIs as we're refactoring.

iamandrewluca: Should we take into consideration that functions are hoisted to the top automatically?

taion: right this is purely a code style thing

iamandrewluca: Adding the type argument I think will add more logic to the function, I splitted better in 2 functions.

iamandrewluca:
Marked as draft because have some questions that appearered after looking at method `onTransitionEnd`.

iamandrewluca:
Ok, the purpose of this deviates, I'll rename, and adapt to new changes.

src/Transition.js
```diff
@@ -316,26 +309,20 @@ class Transition extends React.Component {
...
       ? this.props.nodeRef.current
       : ReactDOM.findDOMNode(this)
 
-    const doesNotHaveTimeoutOrListener =
-      timeout == null && !this.props.addEndListener
-    if (!node || doesNotHaveTimeoutOrListener) {
-      setTimeout(this.nextCallback, 0)
-      return
-    }
+    if (node && this.props.addEndListener && timeout == null) {
```
iamandrewluca, line 317: @taion why `node` is needed here?

src/Transition.js
```diff
@@ -222,14 +207,19 @@ class Transition extends React.Component {
...
       : [ReactDOM.findDOMNode(this), appearing]
 
-    const timeouts = this.getTimeouts()
-    const enterTimeout = appearing ? timeouts.appear : timeouts.enter
+    const enterTimeout = typeof timeout !== 'number' && timeout ? timeout.enter : timeout
+    const appearTimeout = (typeof timeout !== 'number' && timeout
+      ? timeout.appear
+      : timeout
+    ) || enterTimeout // TODO: remove `enterTimeout` fallback for next major
+    const enterTransitionTimeout = appearing ? appearTimeout : enterTimeout
```
taion, line 221: **issue:** can simplify:

```suggestion
    let enterTimeout;
    if (typeof timeout !== 'number' && timeout) {
      // TODO: Remove appear-to-enter fallback for next major.
      enterTimeout = appearing && timeout.appear !== undefined
        ? timeout.appear
        : timeout.enter;
    } else {
      enterTimeout = timeout;
    }
```

note that, unlike in the previous version, your implementation would fall back from a falsy (but defined) `appear` to `enter`

src/Transition.js
taion: i think this is a legacy thing from the conversion to use `nodeRef`; if there's no associated node, then previously it was impossible to call `addEndListener`, because the first arg was expected to be the `node`.

src/Transition.js
```diff
@@ -316,26 +309,20 @@ class Transition extends React.Component {
...
-    if (this.props.addEndListener) {
       const [maybeNode, maybeNextCallback] = this.props.nodeRef
         ? [this.nextCallback]
         : [node, this.nextCallback]
-      this.props.addEndListener(maybeNode, maybeNextCallback)
-    }
 
-    if (timeout != null) {
-      setTimeout(this.nextCallback, timeout)
+      this.props.addEndListener(maybeNode, maybeNextCallback)
```
taion, line 323: this is not equivalent to the original code. previously, if:

- `node` was defined
- `addEndListener` was defined
- `timeout` was defined

then we would call `addEndListener` _and_ we would call `setTimeout` with `this.nextCallback`.

in this revision, there's no way for both `addEndListener` and `setTimeout` to get called

src/Transition.js
```diff
@@ -483,11 +470,7 @@ Transition.propTypes = {
    *
    * @type {number | { enter?: number, exit?: number, appear?: number }}
    */
-  timeout: (props, ...args) => {
-    let pt = timeoutsShape
-    if (!props.addEndListener) pt = pt.isRequired
-    return pt(props, ...args)
-  },
+  timeout: timeoutsShape,
```
taion, line 473: **question:** why the change here? if we don't have a shim to add an end listener, then we should require a timeout to ensure that the transition end callback fires

taion: the "exit" callbacks, rather

src/Transition.js
iamandrewluca: Is this the right behaviour for both of them to call `nextCallback`?

src/Transition.js
iamandrewluca: Is there any case when `node` may be missing?

taion: might be possible if e.g. the transition is rendering a child that's actually not there, maybe?

src/Transition.js
taion: i think the idea is that, if there's a timeout and an end listener, that the first one that triggers will "win"

Issue #410, opened 2018-09-17
When single item is replaced, elements are inserted in wrong order


williamrjribeiro:
**What is the current behavior?**
When  `TransitionGroup` is used for managing `Transition`s, there's an edge case which results in a bug. Let me try to explain:

1. Imagine a todo list with a max of 3 items. One item per row.
1. Items are added at the bottom so the topmost item is the oldest.
1. The list has already 3 items and a new one is added, the oldest item, at the top, is removed to make room for the new one added at the bottom.

This scenario already works as expected. 
![replace-items](https://user-images.githubusercontent.com/1503499/45633744-7f663480-ba98-11e8-97b5-bbf3dd43cfab.gif)

The edge case is when the list has a max of 1. Notice how the `Replaced` item appears at the top of the list instead of the bottom.
![replace-one-bug](https://user-images.githubusercontent.com/1503499/45633771-96a52200-ba98-11e8-8655-67ac0d6b7e08.gif)

**What is the expected behavior?**
The new item should have appeared at the bottom of the list. In other words, when `TransitionGroup` has only one child and in the next render this child is replaced by a new one, the order in which the children are processed should be reversed.

I think the problem might be on [mergeChildMappings](https://github.com/reactjs/react-transition-group/blob/master/src/utils/ChildMapping.js#L38).

You can try a simplified version of the example [here](https://00rqyo26kn.codesandbox.io/).

**Which versions, and which browser / OS are affected by this issue? Did this work in previous versions?**
I reproduced this bug with `^2.3.1`.

If anyone knows a workaround for this problem, please share. 😉

PR #640, opened 2020-06-11
Fix issues #474 ad #410 - If add new items in correct order

### LINKED ISSUE #474: Entering/Exiting flipped for last item in transition group

When removing/adding items in a transition group:
1. When removing the first item and adding an item to the end, first item stays in place until removed.
1. When removing the last item and adding an item to the end, last item swaps place with the item being added.

When removing the last item, it should retain it's position until completely removed.  Entering and Existing items should not *swap* places while in transition.

Remove the first item, list order stays consistent.  Entering item appears at the end of the list and exiting item retains its position.
Remove the bottom item, entering/exiting items swap places.  Entering item appears above item being exited.
https://codesandbox.io/s/k9xx51p555

Maybe the same bug as: https://github.com/reactjs/react-transition-group/issues/410

nol13:
Fixes #474 and #410 

If there are new keys added, order the pending keys before them

### DIFF
--- a/src/utils/ChildMapping.js
+++ b/src/utils/ChildMapping.js
@@ -48,6 +48,15 @@ export function mergeChildMappings(prev, next) {
   let nextKeysPending = Object.create(null)
 
   let pendingKeys = []
+
+  let firstNewKey
+  for (let nextKey in next) {
+    if (!(nextKey in prev)) {
+      firstNewKey = nextKey
+      break
+    }
+  }
+
   for (let prevKey in prev) {
     if (prevKey in next) {
       if (pendingKeys.length) {
@@ -57,6 +66,11 @@ export function mergeChildMappings(prev, next) {
     } else {
       pendingKeys.push(prevKey)
     }
+  }
+
+  if (firstNewKey && pendingKeys.length) {
+    nextKeysPending[firstNewKey] = pendingKeys
+    pendingKeys = []
   }
 
   let i
@@ -73,7 +87,7 @@ export function mergeChildMappings(prev, next) {
     childMapping[nextKey] = getValueForKey(nextKey)
   }
 
-  // Finally, add the keys which didn't appear before any key in `next`
+  // Finally, add the keys which didn't appear before any key in `next` if there were no new keys added
   for (i = 0; i < pendingKeys.length; i++) {
     childMapping[pendingKeys[i]] = getValueForKey(pendingKeys[i])
   }
--- a/test/CSSTransitionGroup-test.js
+++ b/test/CSSTransitionGroup-test.js
@@ -65,8 +65,8 @@ describe('CSSTransitionGroup', () => {
     );
 
     expect(transitionGroupDiv.childNodes.length).toBe(2);
-    expect(transitionGroupDiv.childNodes[0].id).toBe('two');
-    expect(transitionGroupDiv.childNodes[1].id).toBe('one');
+    expect(transitionGroupDiv.childNodes[0].id).toBe('one');
+    expect(transitionGroupDiv.childNodes[1].id).toBe('two');
 
     jest.runAllTimers();
 
@@ -98,8 +98,8 @@ describe('CSSTransitionGroup', () => {
     );
 
     expect(transitionGroupDiv.childNodes.length).toBe(2);
-    expect(transitionGroupDiv.childNodes[0].id).toBe('two');
-    expect(transitionGroupDiv.childNodes[1].id).toBe('one');
+    expect(transitionGroupDiv.childNodes[0].id).toBe('one');
+    expect(transitionGroupDiv.childNodes[1].id).toBe('two');
   });
 
   it('should switch transitionLeave from false to true', () => {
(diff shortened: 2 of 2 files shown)


taion:
i'm just catching up on this issue now – this changes the existing behavior for a replace transition in cases where order matters, doesn't it? e.g. in the "single item" case, there would now be no way to get the newly-inserted item to come "before" the item that's being replaced?

i wonder if the right thing here is to use the value of the key itself as a fallback ordering mechanism, or something to that effect. otherwise we're opting all users into one behavior or the other, when in practice there's no way to know which is unambiguously correct

nol13:
It would change the existing behavior in the case where all of the existing items are removed. If you think of the single item scenario as just a special case of the last item being removed issue, then it would seem to make sense like this. I know in my use case where we have a list of a bunch of items but are only showing say 25 at a time, when the old items leave you'd want the new ones that enter to fill the missing places to come in from the bottom rather than the top, as the old ones animate away.

So if I exit all 25 current items, the new ones that come after it in the list enter from the bottom. This is what it does currently when new items enter, except that if the last existing item leaves, it never comes across any more existing items to place them before. So if a middle item exits, the new one goes on bottom, but if the last item exits, the new item gets added and then the previously last pending item is added behind it.

But ya there could potentially be some code out there that depends on the old behavior. Should #410 be considered a bug? Like would there be some prop that says, "if all items are replaced, place the new items before the old ones, but if only some of them are replaced, they go in list order"

Could use key as fallback, but has there been any assumption that the keys were ordered until now? Seems like that might break existing code as well.

nol13:
Currently definition of this function from the comment says "This function takes a previous set of keys and a new set of keys and merges them with its best guess of the correct ordering," and this seems like a reasonable guess, but ya I could see where it might be useful to have more control over it.

taion:
right; the problem is that if we go from `[a]` to `[b]`, we don't know whether the animation state should be `[a, b]` or `[b, a]`.

i assume you're actually hitting this problem. what sort of API would you find useful for controlling this?

i wonder if we want something like:

```js
<TransitionGroup.Child key="b" placementHint="prepend">
  <MyComponent />
</TransitionGroup.Child>
```

nol13:
Ya I think that would work splendidly.

Just trying to think, if, for some reason, you replaced an item in the list with 2 new items, the first one append and then 2nd prepend, how to handle it. Go by the first one in the group?

taion:
Like this, I think:

```js
<TransitionGroup ...>
  <TransitionGroup.Child key="a" rangeBehavior="prepend">
    <Foo />
  </TransitionGroup.Child>
  <TransitionGroup.Child key="b" rangeBehavior="append">
    <Foo />
  </TransitionGroup.Child>
</TransitionGroup>
```js

Note that this would only really matter in the case where we remove all keys. Everything with `prepend` goes before the removed keys, while everything with `append` goes after.

nol13:
Should work, I think could probably implement it. 

The "only matters when you remove all keys" part might not apply anymore though, since it could also be used when replacing items in the middle of a list.

taion:
Yeah, that could work too. I suspect in practice you might still need to do _some_ key merging, though... I'd do whatever ends up easier to implement.

nol13:
Yup, if you replaced consecutive keys and wanted to interleave the new ones it still wouldn't be able to. Was thinking there could be a 'replacesKey' prop, or go back to the fallback ordering you mentioned or some other way to directly specify. I personally have no need for it at the moment at least.

So this is updated to add a new boolean prop appendOnReplace to the transitions. If it's not set then it will default to prepend and the behavior should be the same as before without any breaking changes. If there are no replacement items then it will get ignored as well, and only apply when determining whether pending items should go before or after new items.

Could change back to string but the bool seemed to work with only the 2 possible values.

taion:
@jquense @silvenon thoughts on the API here?

silvenon:
I'll check this PR out on Saturday, thanks! 🙂  📅

silvenon:
What a tricky issue! At first I thought that we could add an `allChildKeys` prop to `TransitionGroup` so that we could reach more expected conclusions when mapping children, but that way we could solve only the issue where an array is being sliced, not when an entire list is being replaced with a new one and a myriad of other cases you mentioned. My next thought was to provide a some sort of a prop which takes a function with the old list and the new list, but that would just boil down to `TransitionGroup`'s internals.

This issue seems a little beyond me at the moment, but I'll let you know if anything comes to mind.

nol13:
If it's important to have the full control, was thinking there could be a replacesKey prop that would work as an additional hint to appendOnReplace, where if it's replacing a section of existing keys, it will look if the value for its 'replacesKey' exists in the keys it will replace, and if so use that to determine the exact key that it should go before or after.

Or, simplify it and don't worry about the case where you have some elements that would go before and some after, and make it a top level prop on TransitionGroup.

taion:
Okay, what if instead of trying to be clever here, we just let the user pass in a custom callback that will replace our `mergeChildMappings` is specified?

It's not the best API ever, _but_ it should solve any potential customization use case.

nol13:
Would work. Feel like a top level prop for default direction would solve most people issues without having to re-implement internals themselves, but ya not as flexible. Would be happy to try to implement it either way.

Fwiw for what I was doing it was working great with 25 and 50 items, but of course we wanted to start animating ~100 elements in and out at a time and hit some performance issues in Firefox, so had to switch to a 2 step out then in approach. Did try the fast-dom fork too, but doing a bunch of crazy stuff callbacks and such probably wasn't helping. ;)

taion:
I was thinking that we'd export the "prepend instead of append" variant to satisfy the basic use case, but still expose the lower-level API just in case.

I dunno, just a thought – @silvenon @jquense?

mogzol:
Are there any updates for this? I'm running into animation issues in my app that are caused by this, and currently don't really have any way to fix them. The current PR would work great, but any of the other proposed solutions would be fine too, I just need some way to fix it.

PR #653, opened 2020-07-22
fix(Transition): Don't warn on cross-realm nodes

eps1lon:
`node instanceOf Element` checks return false if the node comes from a different window than where the `Transition` module was initialized.

As far as I can tell we don't care if the node is from a different window. We're only interested in the implemented interface so checking for [`nodeType`](https://developer.mozilla.org/en-US/docs/Web/API/Node/nodeType#Node_type_constants) should be sufficient.

The test passes for the previous and proposed implementation without changing the error message. We could just go with `PropTypes.shape({ nodeType: PropTypes.oneOf([1]).isRequired })` but the error message would not be helpful since it wouldn't mention `Element` at all.

Resolves https://github.com/reactjs/react-transition-group/issues/618#issuecomment-624344177
Resolves https://github.com/reactjs/react-transition-group/pull/619#discussion_r457550929

### DIFF
--- a/src/Transition.js
+++ b/src/Transition.js
@@ -391,9 +391,19 @@ Transition.propTypes = {
    *     [test/CSSTransition-test.js](https://github.com/reactjs/react-transition-group/blob/13435f897b3ab71f6e19d724f145596f5910581c/test/CSSTransition-test.js#L362-L437)).
    */
   nodeRef: PropTypes.shape({
-    current: typeof Element === 'undefined'
-      ? PropTypes.any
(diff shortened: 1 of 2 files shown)


koba04:
@eps1lon I missed this PR for the issue, and this has been fixed at #732. I'm sorry that I couldn't take your PR🙏

eps1lon:
> @eps1lon I missed this PR for the issue, and this has been fixed at #732. I'm sorry that I couldn't take your PR🙏

All good. Thanks for the fix!

PR #732, opened 2021-05-27
Fixed `nodeRef` prop type for cross-frame elements

Andarist:
Demo of the problem (and the fix if you uncomment it) presented here: https://codesandbox.io/s/naughty-dan-zps3f?file=/src/App.js

Simple `instanceof` checks always fail for cross-realm checks as technically each realm gets its own, pristine, environment which includes all global constructors (such as `Element`)

### DIFF
--- a/src/Transition.js
+++ b/src/Transition.js
@@ -391,9 +391,18 @@ Transition.propTypes = {
    *     [test/CSSTransition-test.js](https://github.com/reactjs/react-transition-group/blob/13435f897b3ab71f6e19d724f145596f5910581c/test/CSSTransition-test.js#L362-L437)).
    */
   nodeRef: PropTypes.shape({
-    current: typeof Element === 'undefined'
-      ? PropTypes.any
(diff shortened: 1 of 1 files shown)


koba04 (commented):
@Andarist 
👍 Thank you! This looks good to me. Could you fix ESLint errors to pass the CI?

Andarist:
@koba04 I think it will pass now - if not I will come back to it some time later

koba04 (approved):
👍

koba04:
@Andarist Thank you for your fix!

oliviertassinari:
Yes! Happy to see this long-standing issue fixed.

As a side note, on MUI, we fixed it by testing [x.nodeType !== 1](https://github.com/mui-org/material-ui/blob/7dfdb815e40d79e857f5d2e6a7c1519428bbf058/packages/material-ui-utils/src/HTMLElementType.ts#L19).

PR #737, opened 2021-06-09
chore: use prettier for the code formatter

koba04:
`prettier` is installed but it is not used.
This PR uses `prettier` as the code formatter.
I've applied `prettier` to source code files and added npm-scripts for `prettier`.


koba04:
This won't change any behaviors, so I can merge this safely.

PR #741, opened 2021-06-17
chore: migrate storybook to v6

koba04:
This PR has updated the storybook version from v5 to v6. This only includes minimum works to migrate storybook to v6.
After this PR, the start-up time has become slower than before and I'm not sure why this happens.


koba04:
This is quite slow and is not acceptable, so I have to resolve the issue before merging this.

koba04:
https://github.com/reactjs/react-transition-group/pull/741#issuecomment-864381124 has been solved with v6.3, so I'll merge this.

PR #749, opened 2021-07-25
fix: enter animations with mountOnEnter or unmountOnExit 

### LINKED ISSUE #745: Transition enter animation doesn't work correctly with unmountOnExit

When using the `Transition` component with `unmountOnExit`, the exit animation works correctly, but the enter animation doesn't work. However, when I remove `unmountOnExit`, the enter animation works.

When using the `Transition` component with `unmountOnExit`, I expect the enter animation to work correctly.

> Could you provide a [CodeSandbox](https://codesandbox.io/) demo reproducing the bug?

https://codesandbox.io/s/admiring-fog-lhroc?file=/FadeScaleAlert.js

koba04:
fixes #745 
This fixes enter animations with `mountOnEnter` or `unmounOnExit`. However, this might introduce other issues like unexpected animations, so I have to test this before this gets merged.


koba04:
I've added more stories for CSSTransition and it seems to work fine.
I've found some issues around TransitionGroup and couldn't figure out whether Transition should mount an element with unmountOnExit and without mountOnEnter or not. Currently, it's not mounted.
But I think this is another issue, so I won't fix it in this issue.

koba04:
This introduces an additional render because `performEnter` is no longer batched.

anatolzak:
Hi @koba04.

Thanks again for working on this pull request to fix the bug that I reported.

Is there perhaps a timeline to when this pull request will be merged into the main branch?

koba04:
@anatolzak 
I don't have any timeline to merge this yet because it has a risk to cause other problems caused by an additional render to make this animation happen.
But I'll try to work to be able to merge this.

ThomasEsseul:
Hi, any update on that ? I'm really interested in that fix

koba04:
I'll check it again.

koba04:
I've reworked this to get merged.

stories/CSSTransition.js
```diff
@@ -0,0 +1,51 @@
+import React, { useState } from 'react';
+import { storiesOf } from '@storybook/react';
+
+import StoryFixture from './StoryFixture';
+import Fade from './transitions/CSSFade';
+
+function ToggleFixture({ defaultIn, description, children }) {
+  const [show, setShow] = useState(defaultIn || false);
+  console.log(show);
```
koba04, line 9:
```suggestion
```

koba04:
@eps1lon Could you review this PR?

danantal:
Hi guys

I am also pretty interested in this fix - so would be amazing if it can be moved further :)

koba04:
I think I had enough time to get feedback, so I'll merge this.

PR #754, opened 2021-08-24
Update Transition.js

Flcwl:
The initial status should be `unmount` when  setting `mountOnEnter` as `false`, for lazy load just at first render.


Flcwl:
And I don't know why setting `unmount` but `mountOnEnter` is set to `true`.

Flcwl:
Sorry, the prop name `mountOnEnter` misled me.

PR #770, opened 2021-11-23
Switch from Enzyme to Testing Library

eps1lon:
Should ease testing with React 18


test/setupAfterEnv.js
```diff
@@ -0,0 +1,5 @@
+import { cleanup } from '@testing-library/react/pure';
+
+afterEach(() => {
+  cleanup();
+});
```
eps1lon, line 5: We sometimes reset all modules in a beforeEach an re-require. Therefore we can't use the main entrypoint of `@testing-library/react` since that entrypoint may setup `beforeAll` hooks. So we have to use the `pure` entrypoint and manually cleanup

koba04:
Great work 👍 👍 👍  I'll review this.

koba04 (approved):

koba04:
Thank you for your great work!

PR #772, opened 2021-11-23
Move from TavisCI to GitHub Actions

eps1lon:
Example run: https://github.com/eps1lon/react-transition-group/runs/4298012167?check_suite_focus=true


.github/workflows/ci.yml
```diff
@@ -0,0 +1,29 @@
...
+    strategy:
+      matrix:
+        REACT_DIST: [16, 17]
+
+    steps:
+      - uses: actions/checkout@v2
+      - name: Use Node.js 14
+        uses: actions/setup-node@v2
+        with:
+          node-version: 14
```
koba04, line 22: The current active LTS version of Node is 16, so v16 seems to be better.

https://nodejs.org/en/

```suggestion
      - name: Use Node.js 16
        uses: actions/setup-node@v2
        with:
          node-version: 16
```

.github/workflows/ci.yml
```diff
@@ -0,0 +1,29 @@
...
+          cache: 'npm'
+      - run: yarn
+      - run: yarn add react@${{ matrix.REACT_DIST }} react-dom@${{ matrix.REACT_DIST }}
+      - run: yarn --cwd www
+      - run: yarn test
+      - run: npm run build
+      - run: npm run semantic-release
```
koba04, line 29: I guess this wouldn't work on GitHub Actions because this requires tokens like `NPM_TOKEN` and `GITHUB_TOKEN`, which would be set on TravisCI.

```
    env:
      GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
      NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
```

Unfortunately, I don't have permission to set up `semantic-release`. 
@jquense Could you help us?

koba04 (commented):
@eps1lon Thank you for your work, this is great👍 
To move CI from TravisCI to GitHub Actions, we also have to migrate `semantic-release`, so we need @jquense's help.

.github/workflows/ci.yml
eps1lon: The goal of this PR was to move from Travis to GitHub actions. Travis was running on node 14. I think we should do this in a follow-up.

.github/workflows/ci.yml
eps1lon: Yes, I forgot to mention that an admin needs to setup secrets.

koba04: What about removing the script?
Personally, I think it's good to release a new version manually instead of doing it on CI. I don't have the permission though.

koba04: I'd like to introduce some breaking changes like dropping IE support and removing PropTypes in the future. In that time, I'd like to release the breaking changes as a single major version, not separately.
But we can't do that because this script releases a new version automatically, so I'd like to remove running `semantic-release` on CI.

eps1lon: We can collect breaking changes on an `alpha` branch and merge that once it's ready. `semantic-release` will create an `alpha` release channel from that branch for test. We're doing that for `@testing-library/react` and it has been working great so far.

eps1lon:
Will re-open from this remote. Otherwise we'll not get a run. Once it's merged forks will trigger Actions.

PR #798, opened 2022-02-16
build: update babel-preset-gatsby and sass

koba04:
I've updated some outdated packages.

- I've updated `babel-preset-gatsby` to the latest version to fix errors in local.
- I've replaced `node-sass` with `sass` because of the deprecation
  - https://www.npmjs.com/package/node-sass

### DIFF
--- a/www/package.json
+++ b/www/package.json
@@ -14,7 +14,7 @@
   "license": "MIT",
   "dependencies": {
     "@babel/core": "^7.3.4",
-    "babel-preset-gatsby": "^0.1.8",
+    "babel-preset-gatsby": "^2.7.0",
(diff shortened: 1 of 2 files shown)


eps1lon (approved):
For these types of tests it would be nice if we had deploy previews for PRs. Otherwise reviewers would have to manually checkout a branch to test if it works. 

I'm assuming you tested this locally.

koba04:
@eps1lon Thank you! I completely agree with you. I'll set up the preview environment on Vercel or else as another PR.

koba04:
To use Vercel as a deploy preview environment, we would have to use the Pro Plan. Is it possible to use a team for `reactjs` org for this?

image

PR #800, opened 2022-02-18
ci: run lint and build the webpage scripts

koba04:
This PR makes running `lint` and building the webpage on CI. This can find some issues before getting merged.


eps1lon (approved):
This is a nice start! Woudl it be possible to get actual deploy previews i.e. a version of the web page that is built from the PR branch and publicly available on the web?

koba04:
@eps1lon 
> This is a nice start! Woudl it be possible to get actual deploy previews i.e. a version of the web page that is built from the PR branch and publicly available on the web?

I completely agree with you. I think it would be nice if we could use Vercel as the deploy preview feature because the `reactjs` org has been using Vercel on `reactjs.org`. But I don't have permission for the setting. Do you have any ideas?

https://github.com/reactjs/react-transition-group/pull/798#issuecomment-1041814125

eps1lon:
As far as I remember vercel was pretty noisy. We're using GitHub pages for deploys though. I don't know how you do deploy previews with GitHub pages. Might not be worth dealing with.

PR #809, opened 2022-03-21
refactor: migrate `src/` to TypeScript

koba04:
I've migrated all JS files in `src/` to TypeScript.
This doesn't create `.d.ts` files, so I'll work on it as a subsequent PR.

Some implementations were hard to type. I've put `@ts-expect-error` comments for them and expect them to be fixed later. 
I also won't fix errors by the `propTypes` definitions because I'll remove `propTypes` support in the future release.

This is a starting point for TypeScript migration and I expect the types to get better.

### DIFF
--- a/.size-snapshot.json
+++ b/.size-snapshot.json
@@ -1,12 +1,12 @@
 {
   "lib/dist/react-transition-group.js": {
-    "bundled": 98066,
-    "minified": 26272,
-    "gzipped": 8027
+    "bundled": 101943,
+    "minified": 26331,
+    "gzipped": 8046
   },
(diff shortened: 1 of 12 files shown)


eps1lon:
Can you reduce the diffs of the files that are not properly recognized as movied? It's IMO fine if we use a lot of `any` for now. It's more important that the default git config recognizes these as renames and not remove+add.

koba04:
After this PR has been merged, I'll update `ReplaceTransition` to https://gist.github.com/koba04/5575ded08b7ece5ca5326c797949d653

koba04:
@eps1lon Thank you for your feedback; I've made all changes as renames.

PR #810, opened 2022-03-30
chore: use React v18 as the default target for development

koba04:
React v18 has been released 🎉 
https://reactjs.org/blog/2022/03/29/react-v18.html

So I've change the default target version of React to v18. But we still test with v16 and v17 on CI, so this won't be a breaking change.
To support React v18, I've updated `@testing-library/react` and downgrade `@testing-library/react` for v16 and v17 on CI.


package.json
```diff
@@ -80,7 +80,7 @@
     "@semantic-release/npm": "^7.0.5",
     "@storybook/addon-actions": "^6.3.4",
     "@storybook/react": "^6.3.4",
-    "@testing-library/react": "^12.1.4",
+    "@testing-library/react": "^13.0.0-alpha.6",
```
eps1lon, line 83: Just released alpha.7 so it would be nice to use this repo as a canary ;)

eps1lon (approved):

package.json
koba04: Thank you! I've addressed this at https://github.com/reactjs/react-transition-group/pull/810/commits/62a44b0de82714f7526b12d5d953fd06acda0e70

PR #812, opened 2022-04-05
refactor: apply more strict types to ReplaceTransition

koba04:
This is a follow-up PR of #809.
I've applied more strict types to ReplaceTransion, which couldn't be done at #809 to avoid being treated as add/remove because of many diffs.

### DIFF
--- a/src/ReplaceTransition.tsx
+++ b/src/ReplaceTransition.tsx
@@ -71,12 +71,12 @@ class ReplaceTransition extends React.Component<Props> {
     const {
       children,
       in: inProp,
-      onEnter: _onEnter,
-      onEntering: _onEntering,
-      onEntered: _onEntered,
-      onExit: _onExit,
-      onExiting: _onExiting,
-      onExited: _onExited,
+      onEnter,
+      onEntering,
+      onEntered,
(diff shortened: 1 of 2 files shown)


src/ReplaceTransition.tsx
```diff
@@ -47,15 +68,19 @@ class ReplaceTransition extends React.Component<Props> {
...
-    delete props.onEnter;
-    delete props.onEntering;
-    delete props.onEntered;
-    delete props.onExit;
-    delete props.onExiting;
-    delete props.onExited;
+    const {
+      children,
+      in: inProp,
+      onEnter: _onEnter,
```
eps1lon, line 74: Is the rename here only for ESLint? If so, then we should configure the rule appropriately. I remember there was an option specifically for rest-spread.

koba04: @eps1lon Thank you! I've enabled the `ignoreRestSiblings` option at [d8dcd90](https://github.com/reactjs/react-transition-group/pull/812/commits/d8dcd90548ea035f85ede7f6fa944b4e6627fd93)

eps1lon (approved):

PR #824, opened 2022-05-10
test: setup TypeScript settings

koba04:
This migrates test files to TypeScript. This is not necessary, but it's useful to prove whether the TypeScript definition has any problem or not.


test/ChildMapping-test.tsx
```diff
@@ -1,12 +1,7 @@
-let React;
-let ChildMapping;
+import React from 'react';
```
eps1lon, line 1: Not really a fan of TypeScript in runtime tests precisely because runtime tests are annoying to type. The previous test did work differently by clearing modules for each test. A static import is not the same.

If you want to test the types, we should have separate test files for them.

test/ChildMapping-test.tsx
```diff
@@ -32,102 +27,102 @@ describe('ChildMapping', () => {
 
   it('should support mergeChildMappings for adding keys', () => {
     let prev = {
-      one: true,
-      two: true,
+      one: 'one',
```
eps1lon, line 30: This looks like runtime behavior changed. But the PR title says it's only about test. Could you clarify why this change here is necessary?

koba04: The type is `ReactChild`, which doesn't accept `boolean`, so I've changed the test data to String.

eps1lon: Sure but that's not a reason to change runtime behavior, no? So far this looks like a breaking change just so that our tests are type-checked?

test/ChildMapping-test.tsx
eps1lon: >  I prefer not to mock or mutate modules as possible. So it would be nice to use static imports as default.

We're doing neither here. We're resetting the module which is not yet supported with Jest and ESM. So until then, we need to re-require.

koba04: Thank you, I have two questions.
- Why do we need to reset the module even though we haven't used mock or mutating modules?
-  Can't we reset modules with Jest even if we transpile ESM to CJS by `tsc`?

test/ChildMapping-test.tsx
koba04: What does "runtime behavior" mean? This doesn't change any runtime behavior. The TypeScript migration causes this change. Do you mean that we shouldn't do any incompatible changes with DefinitelyTyped?

Issue #834, opened 2022-06-25
Error when running npm install


mkelley33:
> What is the current behavior?

```npm ERR! code ERESOLVE
npm ERR! ERESOLVE unable to resolve dependency tree
npm ERR!
npm ERR! While resolving: react-transition-group@4.4.2
npm ERR! Found: react@18.0.0
npm ERR! node_modules/react
npm ERR!   dev react@"^18.0.0" from the root project
npm ERR!
npm ERR! Could not resolve dependency:
npm ERR! peer react@"^16.8.0" from @restart/hooks@0.3.22
npm ERR! node_modules/@restart/hooks
npm ERR!   dev @restart/hooks@"^0.3.22" from the root project
npm ERR!
npm ERR! Fix the upstream dependency conflict, or retry
npm ERR! this command with --force, or --legacy-peer-deps
npm ERR! to accept an incorrect (and potentially broken) dependency resoluti```

> What is the expected behavior?

That it cleanly installs

> Could you provide a [CodeSandbox](https://codesandbox.io/) demo reproducing the bug?

No, it happens at install time

koba04:
This seems to be caused by the `@restart/hooks` package, not `react-transition-group`. I think you should update `@restart/hooks` to `v0.3.26` or later versions.
https://github.com/react-restart/hooks/blob/master/CHANGELOG.md#0326-2021-01-05

mkelley33:
So I updated to `v0.3.26` of `@restart/hooks`, but it still choked on `storybook/react@6.3.4` so I ran `npm install --legacy-peer-deps` and that seemed to allow it to install just fine.  Then I was able to run `npm run storybook`. @koba04 Should I put in a PR to capture all of this?

Issue #815, opened 2022-04-10
"@types/react-transition-group" depends on "@types/react@18.0.1", so it doesn't work with React 17


Hubro:
> What is the current behavior?

I'm using Next.js with React 17 and Material UI. I have installed `@types/react@17.0.2` manually. However, when I install `@mui/material", it pulls in `@types/react-transition-group`, which in turn pulls in `@types/react@18.0.1`. This doesn't match my React version, and it breaks my build at the type checking stage:

```
./src/components/ProfileBadge.tsx:95:12
Type error: 'Trans' cannot be used as a JSX component.
  Its element type 'ReactElement<any, any> | Component<TransProps, any, any> | null' is not a valid JSX element.
    Type 'Component<TransProps, any, any>' is not assignable to type 'Element | ElementClass | null'.
      Type 'Component<TransProps, any, any>' is not assignable to type 'ElementClass'.
        The types returned by 'render()' are incompatible between these types.
          Type 'React.ReactNode' is not assignable to type 'import("/home/tomas/Dropbox/projects/avesta/fjelldata/client/node_modules/@types/react-transition-group/node_modules/@types/react/index").ReactNode'.

  93 |         >
  94 |           <ExitToAppIcon sx={{ ml: "-10px", mr: "10px" }} />{" "}
> 95 |           <Trans>Log out</Trans>
     |            ^
  96 |         </MenuItem>
  97 |       </Menu>
  98 |     </>
info  - Checking validity of types .npm ERR! code ELIFECYCLE
```

> What is the expected behavior?

I have no idea :confused: I have no idea which package to blame here, but I know `@types/react@18.0.1` should not be installed.

Twathik:
+1

ulrich06:
+1

cq-ubaid-khan:
+1

zadinvit:
If you have older version of npm, just try update npm to version > 8.0.0. 
It worked for me. I had npm version 6.x.x. I tried many solutions, but update npm fix this problem. 
Try check this stackoverflow thread - https://stackoverflow.com/questions/71791347/npm-package-cannot-be-used-as-a-jsx-component-type-errors/71838606#71838606.

cq-ubaid-khan:
[...]

PR #841, opened 2022-07-21
build(peer-deps): add React v17 & v18 support

ivangabriele:
Should fix issues like #834.

The default dev version of React is now v18. So I open it as a draft since I'm not sure yet if we need more tests to ensure this change is valid.

Maybe #815 would force us to break from v16 support?

### DIFF
--- a/package.json
+++ b/package.json
@@ -61,8 +61,8 @@
     ]
   },
   "peerDependencies": {
-    "react": ">=16.6.0",
-    "react-dom": ">=16.6.0"
+    "react": ">=16.6.0 || ^17.0.0 || ^18.0.0",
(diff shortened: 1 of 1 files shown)


ivangabriele:
@koba04 or @eps1lon I know this PR is bare right now but don't hesitate to give me related tasks to make this PR a valid candidate.

koba04:
@ivangabriele 
`>=16.6.0` has already included `^17.0.0` and  `^18.0.0` in its range. Why is this change required?

ivangabriele:
@koba04 My bad indeed! I had an issue with a npm install peer-deps resolution error and I went too fast on this one. `>=` covers it. I close it!

PR #859, opened 2022-10-08
Remove "@1.x" from README

cg-roling:
Looks like 1.x versions require an older version of React, I assume this could be dropped.


cg-roling:
My mistake, I moved too quickly, looks like the React docs just link to the v1 branch of the code:
https://reactjs.org/docs/animation.html
links to:
https://github.com/reactjs/react-transition-group/tree/v1-stable

PR #915, opened 2024-12-24
feat!: support React19 & drop React 16,17 support

koba04:
This only updates CI settings and tests. `react-transition-group` seems to work with React 19 without any updates


ddtch:
someone please pay attention to this PR

jquense:
its in draft and not passing CI @ddtch what do you want us to pay attention to??

@koba04 lmk when this is updated to pass CI

test/Transition-test.js
```diff
@@ -197,35 +196,6 @@ describe('Transition', () => {
     });
   });
 
-  it('should use `React.findDOMNode` when `nodeRef` is not provided', () => {
```
koba04, line 200: better to skip it conditionally based on React version

koba04:
@jquense 
Yeah, CI fails are caused because of old dependencies and the Node version, and `react-transition-group` works with React v19 as long as we use `nodeRef`, so I think this doesn't block anything.

I'd say we should update our build settings , outdated dependencies and documentation system before updating React version.

jquense:
you wanna do it?

koba04:
Yeah, I'll do it when I have time, but it would be helpful someone takes that.

The documentation also depends on PropTypes, which was removed in v19, so it would be better to update the documentation as well.
(Keeping PropTypes won't have any issues, it's just ignored, so this isn't a requirement)

https://react.dev/blog/2024/04/25/react-19-upgrade-guide#removed-deprecated-react-apis

sergioviniciuss:
it would be so helpful if this MR would be included soon🙏

koba04:
@sergioviniciuss 
I think `react-transition-group` works with React v19, do you have any issue that this can fix?

danielstclair:
@koba04 I'm seeing an issue when using the Transition component on React v19:
```
TypeError: react_dom__WEBPACK_IMPORTED_MODULE_3___default(...).findDOMNode is not a function
```

samin15:
> @koba04 I'm seeing an issue when using the Transition component on React v19:
> 
> ```
> TypeError: react_dom__WEBPACK_IMPORTED_MODULE_3___default(...).findDOMNode is not a function
> ```

You need to add a ref and it'll work like this:
`<CSSTransition
                nodeRef={nodeRef}
.
.
</CSSTransition>
`

This will make it not crash but it will still fail on typing through eslint, so you might want to disable it for that.

